PR #11, opened 2019-07-29
add bcrypt support

pilif:
as requested in https://github.com/pilif/sftpgo/commit/32192ce7b8515ef842f292ba6c7923a43235867e#commitcomment-34468602, this adds support to the bcrypt password encoding scheme when authenticating users.

Creating new users still uses Argon2i which is objectively better than bcrypt, but bcrypt is more widely used and thus some sites might still be stuck with bcrypt.

### DIFF
--- a/dataprovider/dataprovider.go
+++ b/dataprovider/dataprovider.go
@@ -22,6 +22,7 @@ const (
 
 	logSender                = "dataProvider"
 	argonPwdPrefix           = "$argon2id$"
+	bcryptPwdPrefix          = "$2a$"
 	manageUsersDisabledError = "please set manage_users to 1 in sftpgo.conf to enable this method"
 	trackQuotaDisabledError  = "please enable track_quota in sftpgo.conf to use this method"
 )
--- a/dataprovider/sqlcommon.go
+++ b/dataprovider/sqlcommon.go
@@ -10,6 +10,8 @@ import (
 	"golang.org/x/crypto/ssh"
 
 	"github.com/alexedwards/argon2id"
+	"golang.org/x/crypto/bcrypt"
+
 	"github.com/drakkan/sftpgo/logger"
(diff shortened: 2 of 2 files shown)


drakkan:
thanks!

drakkan:
Hello,

https://cla-assistant.io/drakkan/sftpgo?pullRequest=11

Thank you!

drakkan:
Hello @pilif 

https://cla-assistant.io/drakkan/sftpgo?pullRequest=11

Thank you!

drakkan:
> as a principle I do not sign CLAs and sure as hell not retroactively.
> 
> Feel free to revert my contribution if that's not acceptable to you.

I think the codeflow is already different in the current version, so signing the CLA is not strictly required.

Just to clarify, the CLA does not change your rights and in fact protects you too. It is useful for the project because after signing the CLA you cannot withdraw permission to use your contribution. Anyway thanks for your reply, appreciated

drakkan:
The CLA also says that "You reserve all right, title, and interest in and to Your Contributions" and, as explained, it protects me as author and the recipients of the software, so the SFTPGo open source project, because it makes clear you cannot withdraw permission to use your contribution.

It is a standard Apache [ICLA](https://www.apache.org/licenses/contributor-agreements.html) not something written by me/us. If you don't agree to sign it is fine. What I need to knnow is what I have to do with this patch.

The code you added basically calls `bcrypt.CompareHashAndPassword` to check a bcrypt password. 
The current code flow is different from the one you contributed in 2019. We still use `bcrypt.CompareHashAndPassword` to check a bcrypt password because there is no other way, but we do several other things before checking the password and call `bcrypt.CompareHashAndPassword`. 
I think I don't have to revert your contribution because it was already rewritten. If you don't agree I'll check again. Thank you

PR #12, opened 2019-07-29
make the server banner configurable

pilif:
as requested in https://github.com/pilif/sftpgo/commit/3ffe4a35fcf9300af82e1f3bcce89702c06b136d#commitcomment-34468605, this commit allows the user to configure the identification string used by the server. 

This is mostly a matter of vanity. I know of no real SSH client that's actually displaying this, but in my case, I felt like showing off :p


drakkan:
thanks!

PR #15, opened 2019-07-31
Support multiple public keys

### LINKED ISSUE #14: Support multiple public keys per account
It would be useful to support multiple public keys per account. Similar to `authorized_keys`.

jovandeginste:
This will parse the public key field as a newline delimited list of public keys.

Return (valid) result on first match.

This should fix #14

### DIFF
--- a/dataprovider/dataprovider.go
+++ b/dataprovider/dataprovider.go
@@ -234,11 +234,14 @@ func validateUser(user *User) error {
 		user.Password = pwd
 	}
 	if len(user.PublicKey) > 0 {
-		_, _, _, _, err := ssh.ParseAuthorizedKey([]byte(user.PublicKey))
-		if err != nil {
-			return err
+		for i, k := range strings.Split(user.PublicKey, "\n") {
+			_, _, _, _, err := ssh.ParseAuthorizedKey([]byte(k))
+			if err != nil {
+				return &ValidationError{err: fmt.Sprintf("Could not parse key nr. %d: %s", i, err)}
+			}
 		}
 	}
+
 	return nil
 }
 
--- a/dataprovider/sqlcommon.go
+++ b/dataprovider/sqlcommon.go
@@ -73,20 +73,21 @@ func sqlCommonValidateUserAndPubKey(username string, pubKey string) (User, error
 		logger.Warn(logSender, "error authenticating user: %v, error: %v", username, err)
 		return user, err
 	}
-	if len(user.PublicKey) > 0 {
-		storedPubKey, _, _, _, err := ssh.ParseAuthorizedKey([]byte(user.PublicKey))
-		if err != nil {
-			logger.Warn(logSender, "error parsing stored public key for user %v: %v", username, err)
-			return user, err
-		}
-		if string(storedPubKey.Marshal()) != pubKey {
-			err = errors.New("Invalid credentials")
-			return user, err
-		}
-	} else {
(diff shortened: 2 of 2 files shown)


drakkan:
Good contribution, thanks!

Probably the user validation in REST API will fail with multiple keys

https://github.com/drakkan/sftpgo/blob/master/dataprovider/dataprovider.go#L236

anyway I can fix this and add test cases and documentation later today and then merge your pull request and push the fix. Thanks!

jovandeginste:
It did not fail in my tests ...

drakkan:
Please try this:

```
curl --verbose -d '{"username":"test_user","home_dir":"/tmp/test_home","permissions":["*"],"public_key":"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQC03jj0D+djk7pxIf/0OhrxrchJTRZklofJ1NoIu4752Sq02mdXmarMVsqJ1cAjV5LBVy3D1F5U6XW4rppkXeVtd04Pxb09ehtH0pRRPaoHHlALiJt8CoMpbKYMA8b3KXPPriGxgGomvtU2T2RMURSwOZbMtpsugfjYSWenyYX+VORYhylWnSXL961LTyC21ehd6d6QnW9G7E5hYMITMY9TuQZz3bROYzXiTsgN0+g6Hn7exFQp50p45StUMfV/SftCMdCxlxuyGny2CrN/vfjO7xxOo2uv7q1qm10Q46KPWJQv+pgZ/OfL+EDjy07n5QVSKHlbx+2nT4Q0EgOSQaCTYwn3YjtABfIxWwgAFdyj6YlPulCL22qU4MYhDcA6PSBwDdf8hvxBfvsiHdM+JcSHvv8/VeJhk6CmnZxGY0fxBupov27z3yEO8nAg8k+6PaUiW1MSUfuGMF/ktB8LOstXsEPXSszuyXiOv4DaryOXUiSn7bmRqKcEFlJusO6aZP0= nicola@p1\ninvalid"}' "http://127.0.0.1:8080/api/v1/user"
```

this does not fail but probably only the first key is validated and I would like to avoid to store invalid public key, at least if the user is saved using the REST API

jovandeginste:
I see what you mean now! You are right, of course!

jovandeginste:
I amended my commit to include validation. This was the output from your curl:

```json
{
  "error": "Validation error: Could not parse key nr. 1: ssh: no key found",
  "message": "",
  "status": 400
}
```

drakkan:
Thanks!

drakkan:
I'm doing some more tests with multiple public keys and even if I don't like much the actual code:

1) the canonical way, in json, to send multiple values is using an array
2) if you use a json library to encode a newline separated string you most likey will get the `\n` escaped  string and not newline.
3) Our actual code will validate only the first key if we get multiple keys separated by an escaped `\n` and it not fails. In this case only the first key will work
4) To fix our code we should split by `'\n'` instead of `"\n"` but this doesn't work the way around

I'm sure that we can fix our code someway and error out if we don't receive a valid newline in public keys string but to keep the code more cleaner and to cause less trouble to the JSON API users I want to change public_key from a string to a string array. 

I know that this is a breaking change ...

drakkan:
I pushed this change with a compatibility layer

https://github.com/drakkan/sftpgo/blob/master/dataprovider/sqlcommon.go#L285

so it should not break your current setup

PR #17, opened 2019-07-31
When path starts with slash, consider it absolute

jovandeginste:
When SQLite path starts with a `/`, we consider this to be an absolute path.

Eg.:

```json
{
  "data_provider":{
    "name":"/var/lib/sftpgo/sftpgo.db"
  }
}
```

(This should maybe be reflected in the documentation?)


drakkan:
Good idea! But I think your patch will not work on Windows, can you please use `filepath.IsAbs` and update the README too? Thanks!

jovandeginste:
Changed the PR. You are free to squash-merge when necessary :-)

drakkan:
thanks!

PR #19, opened 2019-08-01
Support multiple private (host) keys

jovandeginste:
With this patch, one can configure one or more private (host) keys in
the configuration file.

I made it a nested struct, so we can add more options later. Eg. host
certificates might be a useful addition if I can figure out how this is
done in golang's crypto/ssh...


jovandeginste:
(note that I reformated the json parts and that my editor stripped some trailing whitespaces)

drakkan:
Another good contributions, thanks!

Tests now fail since id_rsa does not exist and the configuration contain keys tha don't exist too, so the sftp server does not start.

The default configuration for keys should be empty and maybe the sftp server should start, printing a warning in log file and to the console, if at least one key is valid: for example if two keys are configured and one does not exist or is invalid (ParsePrivateKey fails) while the other one is ok. I'm not completly sure about this, what do you think about?

For host keys maybe we should refuse to use a file if it is group/world-accessible as openssh, but this can be a separate commit.

jovandeginste:
Was this because the tests use `sftpgo.conf`? I can't seem to run the tests myself... I just removed the `keys` from the `sftpgo.conf`.

drakkan:
yes test cases use the default configuration and I think it should work out of the box

drakkan:
I'm a bit busy at the moment I'll give a look at this test failure after work hours. Maybe there is some wrong assumption in sftpd test cases

drakkan:
I did some mistakes modifying this pull request to fix test cases, I create a new branch insteof of modyfing this one. Anaway I merged the change giving to you the patch attribution.

The test failure was for a bug in your path. The test case pass ".." as configDir,

the patch first get the correct path here:

https://github.com/jovandeginste/sftpgo/blob/multiple-keys/sftpd/server.go#L88

so the path is "../id_rsa" but filepath join here:

https://github.com/jovandeginste/sftpgo/blob/multiple-keys/sftpd/server.go#L105

will make the path "../../id_rsa"

jovandeginste:
I see, because the configDir is not absolute. There should be a fix for that too!

drakkan:
fixed in #22 I tryed to modify your pull request but I ended up with a new one,  anyway you are the author for the commit

PR #21, opened 2019-08-01
Use kingpin for flag parsing

jovandeginste:
This is the minimal change to switch to kingin. Current output of `--help`:

```text
$ ./sftpgo --help
usage: sftpgo [<flags>]

Flags:
  --help                        Show context-sensitive help (also try --help-long and --help-man).
  --config-dir="."              Location for SFTPGo config dir. It must contain sftpgo.conf, the private key
                                for the SFTP server (id_rsa file) and the SQLite database if you use SQLite
                                as data provider. The server private key will be autogenerated if the user
                                that executes SFTPGo has write access to the config-dir
  --log-file-path="sftpgo.log"  Location for the log file
  --log-max-size=10             Maximum size in megabytes of the log file before it gets rotated.
  --log-max-backups=5           Maximum number of old log files to retain
  --log-max-age=28              Maximum number of days to retain old log files
  --log-compress                Determine if the rotated log files should be compressed using gzip
  --log-verbose                 Enable verbose logs
```

By default, kingpin tries to mimick the OS's default behaviour, by using `--long-flag` instead of `-long-flag`. This means that at this time, the command line options would be `--config-dir` etc. instead of `-config-dir`. I would argue this is the better way.

Some extra features now:

* `--help-man`: generate man-page
* `--completion-script-bash`: generate bash completion script

I would suggest some improvements after this PR, such as adding short flags.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,8 @@ module github.com/drakkan/sftpgo
 go 1.12
 
 require (
+	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
+	github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4 // indirect
 	github.com/alexedwards/argon2id v0.0.0-20190612080829-01a59b2b8802
 	github.com/fzipp/gocyclo v0.0.0-20150627053110-6acd4345c835 // indirect
 	github.com/go-chi/chi v4.0.2+incompatible
@@ -15,5 +17,6 @@ require (
 	github.com/rs/zerolog v1.14.3
 	github.com/stretchr/testify v1.3.0 // indirect
 	golang.org/x/crypto v0.0.0-20190701094942-4def268fd1a4
+	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
 )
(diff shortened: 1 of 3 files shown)


drakkan:
This could create issues for existing users. I don't really know how many users sftpgo has

jovandeginste:
I don't see a way around this with kingpin...

drakkan:
I documented the --variant so new users will use this way, let's wait a bit before merging. Meantime I'll try to use kingpin to understand it better :)

mcdappdev:
This is a breaking change, right? When are you planning on releasing a stable 1.0?

jovandeginste:
@drakkan I didn't use any extra features of kingpin yet - just wanted the minimal conversion now... It might however help with #23 

@mcdappdev this would be a breaking change

drakkan:
@mcdappdev the argument parser already accept -- syntax.

For example

`sftp -config-dir <path>` and `sftp --config-dir <path>` both already work. 

I updated the docs so new users will use the two minus variant and if we change argument parser in future this will not break their setup. If you use the single minus variant please update your startup script.

Regarding the first stable release I think some feature are still missing, for example:

1) scp support. I don't know if we'll be able to make this work on windows
2) atomic uploads (see Pure-FTPd)
3) we need to take a decision about the command line parser
4) more testing in production setups

We can also add some of these features after 1.0 but we need to wait a bit for reports from production setups (for example a particular sftp client that does not work and similar things). I realeased SFTPGo as public project only two weeks ago I think very few people already use it in production :)

mcdappdev:
Makes sense! Thanks for the response. I just switched our setup over to SFTPGo and it's been running in production for the last few days, it works perfectly. Thanks for such a great library :)

jovandeginste:
We're also using SFTPGo in our environment, and will be using it on production servers in short term.

drakkan:
Good to know!

The last weekend I added atomic uploads support.
SCP support is not so easy so I'll will probably postpone it after the first release.

Now the bad news: I want to try to switch to a cli parser library such as cobra or kingpin before the first release and probably add an integrated CLI too. 

This will be a breaking change, sorry, but I think is better to take these decisions before the first "stable" release.

jovandeginste:
People should pin the version they use anyway? It wouldn't be an issue for us, since I build RPM's from your code (adding some extra stuff like the man page). When I rebase your code, I make sure to update all depending stuff... (And I update on a test server first, of course ;-))

drakkan:
thanks for your contributions but we switched to cobra to have better integration with viper

PR #27, opened 2019-08-05
Add Environment & EnvironmentFile directives

jovandeginste:
Since we can now override settings through environment variables, it's handy to contain those changes in a separate file. This way, sysadmins can generate such a file without having to touch the systemd file.

Meanwhile the current flags are converted to `Environment` entries, which can be overridden in the `EnvironmentFile` if so desired.

If the file is not present, the directive [is ignored](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=)

### DIFF
--- a/init/sftpgo.service
+++ b/init/sftpgo.service
@@ -7,7 +7,10 @@ User=root
 Group=root
 Type=simple
 WorkingDirectory=/etc/sftpgo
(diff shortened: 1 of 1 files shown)


drakkan:
thanks!

PR #39, opened 2019-09-05
Docker alpine multiuser

binou-31:
Sorry for double pull request. I rename the branch `docker-alpine-sftpgo` to `docker-alpine-multiuser` and add fix in the code; 
i'm not an expert of Git :(

Fix : 
- set +x permissions on entrypoint script
- add `ENTRYPOINT` in Dockerfile

I'll delete my older branch with the bad name


drakkan:
No problems, if you do more changes all you need to do is commit them to your local branch and then push to your remote branch, they will be automatically added to your pull request. Thanks again for this pull request.

docker/alpine/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.12-alpine3.10 as builder
+
+RUN apk add --no-cache git gcc g++ ca-certificates \
+  && go get -u github.com/drakkan/sftpgo
+
```
drakkan: try to embed the build info inside the executable and allow to get a stable version too, for example
```
RUN go get -d github.com/drakkan/sftpgo
WORKDIR /go/src/github.com/drakkan/sftpgo
# uncomment the next line to get the latest git version
#RUN git checkout `git rev-list --tags --max-count=1` 
RUN go build -i -ldflags "-s -w -X github.com/drakkan/sftpgo/utils.commit=`git describe --always --dirty` -X github.com/drakkan/sftpgo/utils.date=`date -u +%FT%TZ`" -o sftpgo
```

docker/alpine/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+FROM alpine:3.10
+
+RUN  apk add --no-cache ca-certificates su-exec \
+  && mkdir -p /data /etc/sftpgo
+
```
drakkan: copy a default config inside the provider and allow the user to override it. sftpgo will search config files in the following folders:

1) config dir
2) $HOME/.config/sftpgo
3) /etc/sftpgo
4) current directory

if no config file is found, in your case if the user put no config inside the exposed volume, env var override does not work

docker/alpine/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+COPY --from=builder /go/bin/sftpgo /bin/
+COPY docker-entrypoint.sh /bin/entrypoint.sh
+RUN chmod +x /bin/entrypoint.sh
+
+VOLUME /data
```
drakkan: probably this should include the config dir too

```
VOLUME ["/data", "/path/to/config_dir"]
```

docker/alpine/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+The volume for the configuration is not mandatory, but it will be necessary to configure SFTPgo with environment variables.
+
+### Example
+> 1003 is a custom uid:gid for this instance of SFTPgo
+```
+# Prereq on docker host
+sudo groupadd -g 1003 sftpgrp && \
+  sudo useradd -u 1003 -g 1003 sftpuser -d /home/sftpuser/ && \
+  sudo -u sftpuser mkdir /home/sftpuser/{conf,data} && \
```
drakkan: Creating user and group is a clean way but not strictly required, you could at least document that 

```
chown 1003:1003 <folder>
```
works too.

init/sftpgo-docker.service
```diff
@@ -0,0 +1,29 @@
...
+[Service]
+User=root
+Group=root
+WorkingDirectory=/etc/sftpgo
+Environment=PUID=1003
+Environment=GUID=1003
+EnvironmentFile=-/etc/sysconfig/sftpgo.conf
+ExecStartPre=-docker kill sftpgo
+ExecStartPre=-docker rm sftpgo
+ExecStart=docker run --name sftpgo \
```
drakkan: This way systemd will monitor the docker client and not the container. Anyway we can keep this service file but inside the docker/alpine folder since it is specific for this Dockerfile, for example it cannot be used with the debian Dockerfile I'm working on. Another option is to copy and paste it inside the README as example

binou-31: Yes yes, this init is only for docker with alpine multi-user

docker/alpine/README.md
binou-31: It's just an example for the comprehension. Yes, username and group name are not mandatory

docker/alpine/Dockerfile
binou-31: My first DockerFile included the direcotry of config but, config directory is not required as a volume because we can configure sftpgo with en file.
I don't know if it's a good plan to include and persist default configuration in image if someone want configure product with env values

docker/alpine/Dockerfile
binou-31: see my comment below

docker/alpine/Dockerfile
binou-31: yes, it's fixed

docker/alpine/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM golang:1.12-alpine3.10 as builder
+
```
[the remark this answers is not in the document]
binou-31: Fixed

docker/alpine/Dockerfile
binou-31: Ohh .. I just got it, correct me if I'm wrong. Is it necessary to configure a configuration file for the environment variables to work?

drakkan: Yes, without a config file env vars will not work

drakkan:
I still see some minor issues here, for example, if a user mount a volume on /etc/sftpgo with no configuration file inside it will be no config at all and sftpgo cannot be configured. Anyway I'll merge as is and eventually I'll fix this later.

Thanks for your contribution!

PR #40, opened 2019-09-05
Add quotes to certain strings

jovandeginste:
This will add extra quotes to some strings (or whatever else the variable might contain)


jovandeginste:
Should I care about the above checks?

drakkan:
thanks!

the failed check say that you modified some code not covered by test cases and it remained uncovered after your changes. So no need to worry about, I have not yet found a way to trigger some errors and so add a test case for them

PR #41, opened 2019-09-05
Add connectionID to as many entries as possible

jovandeginste:
I tried to add connectionID to any default logging - not to console logs


jovandeginste:
I need to fix the tests

PR #45, opened 2019-09-06
Refactor the logging system

jovandeginste:
* created a "Log" function for type "Connection"
* created a "log" function for type "Provider"
* replace logger calls to Log/log where possible

I also renamed PGSSQL to PGSQL, as this seemed to be a typo

### DIFF
--- a/dataprovider/bolt.go
+++ b/dataprovider/bolt.go
@@ -8,7 +8,6 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/drakkan/sftpgo/logger"
 	"github.com/drakkan/sftpgo/utils"
 	bolt "go.etcd.io/bbolt"
 )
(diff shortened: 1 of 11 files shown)


jovandeginste:
I'm sorry for the size of this PR ...

drakkan:
thanks, I'll look at this PR in more detail later, as first impression I don't like all this hardcoded strings:

```
provider.log("debug" ...
```

I'll reuse zerolog log level or I'll add some constants for these string, so type errors could not happen

jovandeginste:
@drakkan I changed the strings to consts; probably not what you want in the end, but might make converting easier later.

drakkan:
thanks!

drakkan:
My last commit should not change the generated logs, please let's me know if I'm wrong and thanks for your contribution it is appreciated

jovandeginste:
I was not entirely happy with the way I changed the dataprovider; I think your way is better...

jovandeginste:
(did not test it yet, btw)

PR #49, opened 2019-09-15
Added sha512crypt support

yeonsh:
This adds support to the sha512crypt password encoding scheme when authenticating users.

### DIFF
--- a/dataprovider/dataprovider.go
+++ b/dataprovider/dataprovider.go
@@ -25,6 +25,7 @@ import (
 	"github.com/drakkan/sftpgo/logger"
 	"github.com/drakkan/sftpgo/metrics"
 	"github.com/drakkan/sftpgo/utils"
+	sha512crypt "github.com/nathanaelle/password"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


drakkan:
Hi, thanks for your contribution.
The patch is "technically" fine, maybe only a test case is missing but i can add it myself after merging.

I'm not sure that using sha512 for storing passwords is a good idea nowaday: sha512 is a fast cryptographic hash function and I think it is not specifically designed for storing passwords.

Please take a look at the following links to better understand what I mean or google for "password hashing algorithms":

https://stackoverflow.com/questions/11624372/best-practice-for-hashing-passwords-sha256-or-sha512
https://security.stackexchange.com/questions/52041/is-using-sha-512-for-storing-passwords-tolerable
https://rietta.com/blog/bcrypt-not-sha-for-passwords/
https://crackstation.net/hashing-security.htm
https://security.blogoverflow.com/2013/09/about-secure-password-hashing/

Can you please better explain your use case? We use argon2id by default (currently the most secure algo for passwords) and we support bcrypt and pbkdf2 (both sha256 and sha512 variant), can you please explain why they are not good for you? Thanks!

drakkan:
This is sha512 crypt and not sha512, I'll reconsider it. Anyway please explain your use case and why you cannot use the supported algos that should be more secure anyway. Thanks

jovandeginste:
The use case I can think of, is migrating existing set-ups where you have the (current) hash but not the actual password.

drakkan:
Yes, this is useful for migrating from system user's accounts

drakkan:
Hello,

https://cla-assistant.io/drakkan/sftpgo?pullRequest=49

The contributed code has been rewritten but we would still greatly appreciate it if you could sign the CLA.
Thank you!

PR #50, opened 2019-09-20
Log to journald

jovandeginste:
By default on systems with systemd, send logs to stdout and thus to journald.


drakkan:
Thanks!

drakkan:
Updated release tar too

PR #74, opened 2020-02-14
portable: set data_provider.name to empty

### LINKED ISSUE #73: portable mode sftpgo.db error
I build `sftpgo` docker image with alpine instructions.

I am trying to start server with `sftpgo portable`, but it throws `ERR error initializing data provider: stat sftpgo.db: no such file or directory` error all the time. 

I investigated the code and portable mode selects `memory` data provider. Unfortunately I could not find the bug.

enescakir:
Default `data_provider.name` is `sftpgo.db`. Even if driver is `memory`, it checks existence of `sftpgo.db` file.

I set `data_provider.name` to empty string when start portable mode.

Fixes #73

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -150,6 +150,7 @@ func (s *Service) StartPortableMode(sftpdPort int, enabledSSHCommands []string, 
 	}
 	dataProviderConf := config.GetProviderConf()
 	dataProviderConf.Driver = dataprovider.MemoryDataProviderName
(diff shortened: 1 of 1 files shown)


drakkan:
You are right, it was broken by this commit:

https://github.com/drakkan/sftpgo/commit/bcaf283c35c98309673a473210bd371806677c82

thanks

PR #76, opened 2020-02-15
s3: enable shared config state

enescakir:
If `sftpgo` is running on an Amazon EC2 instance, it needs to use IAM roles.
It doesn't have explicit access token on EC2 instance. 

`AWS SDK for Go` has ability to use shared config state. It's recommend method for EC2 instances on `Specifying Credentials` section of [documentation](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -530,7 +530,12 @@ SFTPGo uses multipart uploads and parallel downloads for storing and retrieving 
 
 The configured bucket must exist.
 
-To connect SFTPGo to AWS you need [Access Keys](https://docs.aws.amazon.com/general/latest/gr/aws-sec-cred-types.html#access-keys-and-secret-access-keys), a `region` is required too, here is the list of available [AWS regions](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions). For example if your bucket is at `Frankfurt` you have to set the region to `eu-central-1`. You can specify an AWS [storage class](https://docs.aws.amazon.com/AmazonS3/latest/dev/storage-class-intro.html) too, leave blank to use the default AWS storage class. An endpoint is required if you are connecting to a Compatible AWS Storage such as [MinIO](https://min.io/).
+To connect SFTPGo to AWS you need to specify credentials, and a `region` is required too, here is the list of available [AWS regions](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions). For example if your bucket is at `Frankfurt` you have to set the region to `eu-central-1`. You can specify an AWS [storage class](https://docs.aws.amazon.com/AmazonS3/latest/dev/storage-class-intro.html) too, leave blank to use the default AWS storage class. An endpoint is required if you are connecting to a Compatible AWS Storage such as [MinIO](https://min.io/).
+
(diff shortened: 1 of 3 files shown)


drakkan:
Thanks!

drakkan:
Hi @enescakir 

can you please check that this commit does not break your use case?

https://github.com/drakkan/sftpgo/commit/58253968fcef4432abdcccf8b412b346032dbbd2

this fix the case when you update a user and you set empty s3 key and secret if they were previously not empty.

I don't have a EC2 instance to test there

enescakir:
It looks like OK. If I get further errors, I will let you know.

drakkan:
Great, thanks for your feedback. 

For aws there is this know bug:

https://github.com/drakkan/sftpgo/blob/master/vfs/s3fs.go#L218

This could be a trivial question, I implemented AWS support because there are many compatible objects storages that implements AWS protocol but I don't use it myself. 
AWS itself provides managed SFTP transfers that use S3 as backend, out of curiosity why do you prefer SFTPGo (that needs to be installed somewhere) to this service? Is it very expensive? I would like to better understand this use case, thanks!

enescakir:
The known bug is not a problem for us. 
Our customer send some csv files to us via SFTP. It doesn't have heavy traffic. Solution like `SFTPGo` is enough for us. It supports S3 and logs user actions.

Unfortunately, `AWS Transfer for SFTP` is very expensive. It's $0.30 per hour without data transfer charges. Even if we don't transfer any data, it costs $216 monthly. It's not feasible for our use case.

The basic AWS EC2 `t3.micro` instance (2vCPU, 1 GB memory) is $0.012 per hour. It costs $8.46 per month and it's performance enough for us.

I have question for you. :) You answer questions and solve problems very quickly. Is `SFTPGo` your side project? Or do you use `SFTPGo` in your business?

enescakir:
I can add some AWS EC2 deployment tutorials for `SFTPGo` to this repository, It might increase usage of `SFTPGo`.

In addition, you can put official images to Docker Hub. It makes deployment easier.

drakkan:
> The known bug is not a problem for us.
> Our customer send some csv files to us via SFTP. It doesn't have heavy traffic. Solution like `SFTPGo` is enough for us. It supports S3 and logs user actions.
> 
> Unfortunately, `AWS Transfer for SFTP` is very expensive. It's $0.30 per hour without data transfer charges. Even if we don't transfer any data, it costs $216 monthly. It's not feasible for our use case.
> 
> The basic AWS EC2 `t3.micro` instance (2vCPU, 1 GB memory) is $0.012 per hour. It costs $8.46 per month and it's performance enough for us.
> 

thanks for the explaination

> I have question for you. :) You answer questions and solve problems very quickly. Is `SFTPGo` your side project? Or do you use `SFTPGo` in your business?

I work on SFTPGo in my free time, I don't use it in production yet and I have only one potential customer that could be interested to it, but it should migrate from FTP to SFTP and this could never happen :)

drakkan:
> I can add some AWS EC2 deployment tutorials for `SFTPGo` to this repository, It might increase usage of `SFTPGo`.
> 

this would be great

> In addition, you can put official images to Docker Hub. It makes deployment easier.

I'm not a big docker fan anyway I'm already evaluating this idea, I'll probably add SFTPGo to DockerHub sometime in future

PR #77, opened 2020-02-20
[WIP] Add support for dynamic, semi-virtual filesystem

jovandeginste:
This should be considered work-in-progress.

We have a special use case where users have access to a set of
locations, which varies over time (at a potentially fast pace).

As an example, a user "user1" can start services within our container
infrastructure. In the service definitions, (s)he will define a number
of volumes (in Docker notation), eg.:

```
--name service1-frontend
-v /nfs/volume/service1/data:/data
-v /nfs/volume/service1/config:/config
```

We have a microservice which parses these job definitions and will emit
a json array per user through a REST endpoint:

```
GET /volumes_for/user1
```

```json
[
  "/service1-frontend/data:/nfs/volume/service1/data:",
  "/service1-frontend/config:/nfs/volume/service1/config:"
]
```

With this code in `sftpgo`, I can configure a user "user1", select the
new Storage type "Dynamic", add <https://example.com/volumes_for/user1>
as the "Dynamic configuration source URL".

Then, when "user1" authenticates through sftp, (s)he will see the
following directory structure:

- `service1-frontend/`
  - `data/` -> contains data from `/nfs/volume/service1/data`
  - `config/` -> contains data from `/nfs/volume/service1/config`

(Obviously, this will add more services and folders within those
services, depending on the json response.)

This enables our "user1" to access the files which are inside the
container through SFTP, eg. to get log files, debug input files, etc.

For now, only browsing the directory structure and downloading files
works.

### DIFF
--- a/dataprovider/dataprovider.go
+++ b/dataprovider/dataprovider.go
@@ -606,6 +606,10 @@ func saveGCSCredentials(user *User) error {
 }
 
 func validateFilesystemConfig(user *User) error {
+	if user.FsConfig.Provider == 2 {
+		user.FsConfig.S3Config = vfs.S3FsConfig{}
+		return nil
+	}
 	if user.FsConfig.Provider == 1 {
 		err := vfs.ValidateS3FsConfig(&user.FsConfig.S3Config)
 		if err != nil {
--- a/dataprovider/user.go
+++ b/dataprovider/user.go
@@ -65,10 +65,14 @@ type UserFilters struct {
 
 // Filesystem defines cloud storage filesystem details
 type Filesystem struct {
-	// 0 local filesystem, 1 Amazon S3 compatible, 2 Google Cloud Storage
-	Provider  int             `json:"provider"`
-	S3Config  vfs.S3FsConfig  `json:"s3config,omitempty"`
+	// 0 local filesystem
+	Provider int `json:"provider"`
+	// 1 Amazon S3 compatible
+	S3Config vfs.S3FsConfig `json:"s3config,omitempty"`
+	// 2 Google Cloud Storage
 	GCSConfig vfs.GCSFsConfig `json:"gcsconfig,omitempty"`
+	// 3 Dynamic Configuration from URL
+	DynamicConfigurationURL vfs.DynFsConfigURL `json:"dynamic_configuration_url,omitempty"`
 }
 
 // User defines an SFTP user
@@ -122,6 +126,9 @@ type User struct {
 
 // GetFilesystem returns the filesystem for this user
 func (u *User) GetFilesystem(connectionID string) (vfs.Fs, error) {
+	if u.FsConfig.Provider == 2 {
+		return vfs.NewDynFs(connectionID, u.GetHomeDir(), u.FsConfig.DynamicConfigurationURL)
+	}
 	if u.FsConfig.Provider == 1 {
 		return vfs.NewS3Fs(connectionID, u.GetHomeDir(), u.FsConfig.S3Config)
 	} else if u.FsConfig.Provider == 2 {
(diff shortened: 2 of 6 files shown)


drakkan:
@jovandeginste thanks for your contribution. I like the idea to have virtual folders and this is in my personal TODO too.

I'm not sure we need a new filesystem implementation here: your are using a local filesystem, I'll read again the code and your explaination later but my first impression is this one:

- we could add support for virtual folders that is something that could be useful in general, for example adding a new user configuration directive that allows to specify a list of virtual folders mapping, something like this one:

```
[{"sftp_path":"/virtual1","filesystem_path":"/app/folder1"},...]
```

virtual folders are basically as symlinks, that we currently don't allow if outside the home directory, but they don't need to exist on the filesystem
- we could use external auth to get the dynamic virtual folders mapping
- we could improve external auth adding a mode that allows to update existing users, for example you have this existing user 

```
{"username":"aaaa","home_dir":"/tmp/bb",....}
```

you return from your external auth script something like this (a partial user):

```
{"home_dir":"/tmp/cc"}
```

and this configuration will be merged with the existing user that will become:


```
{"username":"aaaa","home_dir":"/tmp/cc",....}
```

I would like to keep SFTPGo as general as possible and allowing customization using external programs/scripts instead of incorporating very specific code.
We could also add a preauth hook to update user's fields (as script or http call) but it should return an SFTPGo user (or part of it for partial updates) and a foreign structure/list.
 
I don't want to discourage your work and obviously I can change my mind, this is only my first impression.

jovandeginste:
Hi @drakkan 

Just to be clear, this is a limited implementation to basically scratch my own itch. I would be more than happy to have a more generic (an better) implementation, and I trust you to handle this (as you have proven so far).

Some context:
The external auth would be a solution to the extra configuration at this moment. However, I don't want to reinvent the authentication mechanism itself. Only add some extra information. Doable? The idea of extending the user information this way is a good one!
My example of the incoming json is not too far from you suggestion. Mine is more limited, yours is more extendible. Can perfectly live with it.
Finally, please take a look at my code. Not _my_ code per se, but the `vfs` and `mapfs` packages that I use. You will see that I don't have too much code really, most of it is parsing the json :wink: 

So: when you have time, revisit this PR and please provide more thoughts. I have a few days off work, so not sure if I will be programming much in the coming week. Feel free to cherry-pick and/or reimplement the whole thing. You probably understand my use case now, but I didn't want to just give you work without even trying to do my share :smile:

jovandeginste:
Oh, and I found this [afero package](https://github.com/spf13/afero) along the way; maybe it can be useful?

drakkan:
Hi, 

my plan for this is the following:

1) add support for virtual folders as explained above
2) add a pre auth hook (using a script and/or an http request) that let you modify some user fields just before the authentication. The hook must return a (partial) SFTPGo user json serialized and probably it will receive as input the current user json serialized

what do you think about? 

Regarding afero I looked at it in the past when I added S3 support, they don't want to merge S3 support (there is an external fork) and we need a different strategy for creating files on S3/Cloud storages since  we receive the file as chunks via SFTP (I'm using my pipeat fork for this). So instead of forking afero and making changes that probably will never be accepted upstream I added my own implementation that is very specific for sftp transfers and cannot be used as general filesystem abstraction

drakkan:
@jovandeginste please test virtual folders + before login program.

You can execute your custom http requests inside the before login program and modify the user if needed.

I think that we have enough new features to release 0.9.6 so please help me to test these changes, thanks!

jovandeginste:
Wow, you work so fast. As I said, I have a few days off, so I'll test and report back in March...

Issue #69, opened 2020-02-08
Relatively lower performance than OpenSSH


HiFiPhile:
Hi,
Thanks for this great project ! 

I did some test in my environment and the transfer speed is much lower than OpenSSH.

**Server** ||
--- | --- |
OS| Debian 10.2 x64 |
CPU| Ryzen5 3600 |
RAM| 64GB ECC |
Disk| 3* Intel P4510 4TB RAID0 |
Ethernet| Mellanox ConnectX-3 40GbE|

**Client** ||
--- | --- |
OS| Windows 10 1909 x64 |
CPU| Threadripper 1920X |
RAM| 64GB ECC |
Disk| Samsung 960EVO 1TB |
Ethernet| Mellanox ConnectX-3 40GbE|

Under Filezilla I can get 500MB/s with OpenSSH, but only about 200MB/s with sftpgo.

In both case I'm using `AES256-CTR` as cipher and `SHA-256` as MAC, I've also tried `AES128-CTR` but nothing changes.

CPU usage of sftpgo is higher than OpenSSH:
``` 
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND 
 4527 sftp      20   0 1795576  52044   8628 R 133.5   0.6   2:12.13 sftpgo 
```
```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND 
27934 xxxxxx    20   0   17112   5360   4188 R  67.8   0.1   0:10.01 sshd                                                  
27942 xxxxxx    20   0   17112   5344   4176 R  27.4   0.1   0:12.52 sshd 
```

In both case I've got a maximum TCP window size of 4MB.

drakkan:
Hi,

thanks for taking time to do this test, very appreciated.

I don't have a 40 gigabit ethernet card. 

Since the first implementation I noticed, testing on localhost, that sftpgo is a bit slower and uses more CPU than openssh so I can confirm what you report but the difference is not so big in my tests. 

In real environments SFTPGo was able to completly saturate my network card, using more CPU than OpenSSH, but my test were limited to 1 gigabit ethernet cards.

Can you please test multiple parallel transfers and report if you are able to saturate your 40 gigabit network?

Can you repeat the same test using scp? SCP is implemented from scratch inside SFTPGo while for sftp we use pkg/sftp library. If scp will perform better I'll try to see if pkg/sftp can be improved.

[...]

PR #87, opened 2020-03-03
Add performance optimization summary

HiFiPhile:
A simple summary

### DIFF
--- a/README.md
+++ b/README.md
@@ -900,9 +900,40 @@ The **connection failed logs** can be used for integration in tools such as [Fai
 
 SFTPGo can easily saturate a Gigabit connection on low end hardware with no special configuration, and this is generally enough for most use cases.
 
-The main bootlenecks are the encryption and the messages authentication, so if you can use a fast cipher with implicit message authentication, for example `aes128-gcm@openssh.com`, you will get a big performance boost.
+For Multi-Gig connections, some performance improvements and comparisons with OpenSSH are being discussed here [issue](https://github.com/drakkan/sftpgo/issues/69),  some of them need upstream updates so there are not included in the released version yet. To summarize:
+- In current state with all performance improvements applied, SFTP performance is very close to OpenSSH however CPU usage is higher. SCP performance match OpenSSH.
+- The main bottlenecks are the encryption and the messages authentication, so if you can use a fast cipher with implicit message authentication, such as `aes128-gcm@openssh.com`, you will get a big performance boost.
+- SFTPGo's SCP implementation is more efficient than SFTP, in *nix environment it performs better than SFTP. 
+- Load balancing with HAProxy can greatly improve the performance if CPU not become the bottleneck.
 
-There is an open [issue](https://github.com/drakkan/sftpgo/issues/69) with some other suggestions to improve performance and some comparisons against OpenSSH.
+If you compile SFTPGo from source, you can:
+- AES-CTR optimization of golang compiler hasn't been merged yet, you can apply the patch yourself. [patch](https://go-review.googlesource.com/c/go/+/51670)
+- Use minio/sha256-simd to accelerate MAC computation, which could improve the performance by 50% on newer CPU architectures.
+```
+diff --git a/go.mod b/go.mod
+index f1b2caa..a3e2ba5 100644
+--- a/go.mod
++++ b/go.mod
+@@ -43,3 +43,5 @@ require (
+ )
+ 
+ replace github.com/eikenb/pipeat v0.0.0-20190316224601-fb1f3a9aa29f => github.com/drakkan/pipeat v0.0.0-20200123131427-11c048cfc0ec
++
++golang.org/x/crypto => github.com/drakkan/crypto v0.0.0-20200211081002-cc78d71334be
+```
+- A new allocator which greatly improve parallel loads is being tested here.
+```
+diff --git a/go.mod b/go.mod
+index f1b2caa..4a3be8a 100644
+--- a/go.mod
++++ b/go.mod
+@@ -43,3 +43,5 @@ require (
+ )
+ 
+ replace github.com/eikenb/pipeat v0.0.0-20190316224601-fb1f3a9aa29f => github.com/drakkan/pipeat v0.0.0-20200123131427-11c048cfc0ec
++
++replace github.com/pkg/sftp => github.com/drakkan/sftp v0.0.0-20200227085621-6b4abaad1b9a
+```
 
 
 ## Acknowledgements


drakkan:
Thanks, I would remove the part about WinSCP, I don't want to blame WinSCP or any other sw in the docs.

I think would be useful if you can describe your test server and report the results against OpenSSH that is the reference implementation (and is written in C and use OpenSSL for crypto and so it is very hard to beat) do you agree?

There are other small notes too, but I can add them myself, for example I don't think that my crypto branch will improve sha performance if avx512 is available (I should add one or two line of code, but I haven't an AVX512 server available for testing )

HiFiPhile:
Hi,
I made some modification.
`I would remove the part about WinSCP`
I've removed the WinSCP part.

`I think would be useful if you can describe your test server and report the results against OpenSSH that is the reference implementation `
I've add a phrase about the comparaison, however since the README is already very dense, I don't think it's a good idea to put too much details.

`I don't think that my crypto branch will improve sha performance if avx512 is available`
So some modifications are needed to let AVX512 enabled ? Me neither I don't have a AVX512 machine...

drakkan:
> Hi,
> I made some modification.
> `I would remove the part about WinSCP`
> I've removed the WinSCP part.
> 
> `I think would be useful if you can describe your test server and report the results against OpenSSH that is the reference implementation `
> I've add a phrase about the comparaison, however since the README is already very dense, I don't think it's a good idea to put too much details.
> 
> `I don't think that my crypto branch will improve sha performance if avx512 is available`
> So some modifications are needed to let AVX512 enabled ? Me neither I don't have a AVX512 machine...

we probably need to add replace `sha256.New`

https://github.com/drakkan/crypto/blob/master/ssh/mac.go#L50

with something like this to my crypto branch for AVX512

```
server := sha256.NewAvx512Server()
sha256.NewAvx512(server)
```

or eventually enable avx512 directly inside sha256-simd if it improve ssh/sftp performance

jovandeginste:
I would suggest creating a separate performance document, and referencing it from the main readme. Maybe add a summary in the readme... The performance document could then contain a lot of information and reproduction code, without detracting from the important parts...

drakkan:
> I would suggest creating a separate performance document, and referencing it from the main readme. Maybe add a summary in the readme... The performance document could then contain a lot of information and reproduction code, without detracting from the important parts...

I agree we could create a `docs` directory and add there a "Performance.md" doc

HiFiPhile:
Yes we could reorganize documents. I feel currently README.md is too dense for the first look.

`My idea is to close the issue #69 and summarize the collected info, the hw used for the tests is important, as explained above an AVX512 capable hw could be slower`

Maybe we could wait the new allocator is more elaborated before closing it ?

drakkan:
> Yes we could reorganize documents. I feel currently README.md is too dense for the first look.
> 

you are right we should have separate sections

> `My idea is to close the issue #69 and summarize the collected info, the hw used for the tests is important, as explained above an AVX512 capable hw could be slower`
> 
> Maybe we could wait the new allocator become more elaborated before closing it ?

My intention was to avoid to a newcomer to read 30+ comments, 

I suspect that the new allocator will require several months before it could be merged upstream, anyway we have no hurry to close that issue :)

drakkan:
I'll try to reorganize the docs the next weend, I'll remove many details from the main README and I'll add separate documents for specific things in a new `docs` directory. The main README will have links to this specific docs. This way the documentation should be less confused. 

So for example the "Performance" paragraph will become a simple link or it will contain a very concise introduction and then it will contain a link to `/docs/Performance.md` for more details.

Maybe something like an index could help too

@jovandeginste @HiFiPhile @others if you have time your help is welcome, thanks!

jovandeginste:
Should we also refer to an example json config instead of including it verbatim in the readme?

drakkan:
> Should we also refer to an example json config instead of including it verbatim in the readme?

yes, so we don't have to copy it each time a new configuration option is added

jovandeginste:
@drakkan I've done some preliminary changes in a branch [here](https://github.com/jovandeginste/sftpgo/tree/docs)

drakkan:
@jovandeginste it looks very promising, thanks! I think we can compact other sections too, for example the configuration section is huge, we can split the configuration options in a separate doc (all the keys in sftpgo.json), a typical user will start with the defaults and then read the docs only if he want to customize something

anyway great job, thanks!

jovandeginste:
@drakkan sure, I just did some easy stuff :-) Since this will take some time and collaboration, I'm not sure how we should best proceed... I could grant write access to my fork of this project to you and others?

drakkan:
> @drakkan sure, I just did some easy stuff :-) Since this will take some time and collaboration, I'm not sure how we should best proceed... I could grant write access to my fork of this project to you and others?

Yes, this is a way or you could simple send a pull request with your partial work, I think I can find some time the next weekend to try to complete what is missing

drakkan:
@HiFiPhile thanks to the hard work of @jovandeginste we now have a much better docs. I really like it!

Regarding this pull request, here is what I would like to see in the new [performance](https://github.com/drakkan/sftpgo/blob/master/docs/performance.md) section:

- a description of your hardware
- a description of the tested configuration:
  1. baseline configuration: current sftpgo git master that include no golang patch and the current pkg/sftp master. This is the configuration used by anyone
  2. performance optimized configuration with a brief description of the used patch
- tested ciphers, this should include `aes128-gcm@openssh.com` and `aes128-ctr` or `aes256-ctr`
- a brief description of the impact of the applied patches in the test results: the golang patch improve aes-ctr, the allocator patch gives general improvement for pkg/sftp, sha256-simd improves mac performance. So `aes128-gcm@openssh.com` is not improved applying the aes-crt and sha256-simd patch, the numbers for this cipher show the performance gain that can be obtained using the allocator patch (this should clarify the things for pkg/sftp maintainers, see the last comment [here](https://github.com/pkg/sftp/issues/334))
- the numbers for the baseline and patched configuration and a comparison with OpenSSH using the same ciphers, optionally the number for haproxy too. SCP numbers could be useful too: they show what we can achieve with the golang ssh stack. SCP is much simpler than SFTP so it is expected that SFTP is a bit slower

Please let me know what do you think about, 

thanks!

HiFiPhile:
Yeah the docs are much better now ;)

I agree with you with the performance section.

 I'll also try to write a step-by-step guide to run SFTPGo as service on Debian, or somebody wants do it 👍

drakkan:
@HiFiPhile I added some basic instructions to run SFTPGo as service, I tested on a CentOS 7 VM, I think they should work on Debian too. 

The only difference with the Arch package that I maintain is the the SQLite database will be installed in /etc/sftpgo instead of /var/lib/sftpgo, I omitted that step to avoid complications for MySQL and PostgreSQL users

PR #94, opened 2020-03-13
s3: export PartSize parameter

murlock:
By default AWS SDK use part_size of 5 MB. 

For big files, it is not ideal case. 
For Hadoop, it is not uncommon to use 512 MB.

Note:
Editing existing configuration will requires to set this value

### DIFF
--- a/httpd/httpd_test.go
+++ b/httpd/httpd_test.go
@@ -2010,6 +2010,7 @@ func TestWebUserS3Mock(t *testing.T) {
 	user.FsConfig.S3Config.Endpoint = "http://127.0.0.1:9000/path?a=b"
 	user.FsConfig.S3Config.StorageClass = "Standard"
 	user.FsConfig.S3Config.KeyPrefix = "somedir/subdir/"
+	user.FsConfig.S3Config.PartSize = 5
 	form := make(url.Values)
 	form.Set("username", user.Username)
 	form.Set("home_dir", user.HomeDir)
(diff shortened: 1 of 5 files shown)


drakkan:
Hi,

thank you for the contribution, I'll review it later today or during the weekend. 

If this is useful for your use case I have no problem to merge it (there are some failing test cases, the openapi definition and the docs need to be updated ecc.. but I can fix these things myself) but we need to at least add a documentation note since a big part size could create issues if the bandwidth between the sftp server and the sftp client is much bigger than the one between the s3 backend and the sftp server.

Let's me better explain with an example:

- sftp client has a 1000 Mbps connection to the sftp server
- sftp server has 1 Mbps connection to the s3 backend
- using a 5MB partsize and concurrency = 2 the aws sdk will cache at most 10 MB, so when the client ends sending the data to the sftp server it have to wait while the aws sdk sends the latest 10MB. If we set part size = 512 then the client have to potentially wait in idle state while the sftp server uploads the latest 1024MB and it will probably timeouts.

There is no issue if the bandwidth between the s3 backed and the sftp server is greater than the bandwidth between the sftp server and the sftp client.

drakkan:
Thanks! If you are interested to submit other pull requests please take a look at the next commit that adds documentation and test cases.

I also did some minor changes and I'll make configurable upload concurrency to

murlock:
Thanks, I'll check the next commit !

For the record, my company provides a S3 compatible object storage with a forked swift gateway:
When a big file is written with small parts, there is some side effects on gateway:
- it can be long to CompleteMpu
- if data is stored with Erasure Coding (6+3 for example), it means that each read on a part triggers 6 connections to reconstruct data before sending back to client. It implies some latency.

drakkan:
Thanks for the info, I'm glad that SFTPGo is useful for your use case! If you have other questions or feature requests please open issues to discuss about them, thanks!

drakkan:
Hello,

https://cla-assistant.io/drakkan/sftpgo?pullRequest=94

Thank you!

PR #105, opened 2020-04-08
Add a version build arg to the Debian Dockerfile.

henriklundahl:
This adds a build arg to the Debian Dockerfile to be able to build latest master, the latest tag or a specific tag.

### DIFF
--- a/docker/sftpgo/debian/Dockerfile
+++ b/docker/sftpgo/debian/Dockerfile
@@ -3,8 +3,9 @@ FROM golang:latest as buildenv
 LABEL maintainer="nicola.murino@gmail.com"
 RUN go get -d github.com/drakkan/sftpgo
 WORKDIR /go/src/github.com/drakkan/sftpgo
-# uncomment the next line to get the latest stable version instead of the latest git
-#RUN git checkout `git rev-list --tags --max-count=1`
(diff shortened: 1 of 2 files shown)


drakkan:
Thanks, very appreciated. Are you interested to do the same improvement for the Alpine based Dockerfile?

henriklundahl:
Sure, I'll give it a shot.

By the way, why are the Dockerfiles so different?

drakkan:
> Sure, I'll give it a shot.
> 
thanks!

> By the way, why are the Dockerfiles so different?

I wrote the Debian Dockerfile myself, the Alpine one is a user contribution, so they have a different structure. I never used Alpine myself and I'm not even a big Docker fan so I don't know it very well, I basically verified that it builds and runs and I accepted the contribution. Both Dockerfiles should work fine even if they are a bit different, if you want to suggest improvements they are welcome!

drakkan:
Thanks!

This change allow to build a specific commit too, for example

```
docker build -t="drakkan/sftpgo" --build-arg TAG=c93d8cecfce24dfaed131dcdd69012514719b34b .
```

henriklundahl:
Yes, that's true. Even a branch would work.

Thanks for merging!

PR #106, opened 2020-04-09
Add a version build arg to the Alpine Dockerfile.

henriklundahl:
This adds a build arg to the Alpine Dockerfile to be able to build latest master, the latest tag or a specific tag.


drakkan:
Thanks!

PR #111, opened 2020-04-28
add verbose flag for portable mode

enescakir:
I couldn't find a way to disable debug logs on portable mode.

I add a flag for it.


drakkan:
Thans!

PR #112, opened 2020-04-28
add github action workflow for code quality

enescakir:
What do think about adding code quality checkers for new PRs @drakkan?

[golangci-lint](https://github.com/golangci/golangci-lint) is a great tool and contains lots of linters.

It's little bit annoying at the beginning, but I think following conventions is important for open source projects. 

We can enable/disable some linters or specific checks. It's fully configurable.

This PR shows example usage with Github Actions.

[Example Run](https://github.com/enescakir/sftpgo/runs/626163373?check_suite_focus=true)

golangci-lint

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -29,4 +29,6 @@ jobs:
       run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.24.0
 
     - name: Run golangci-lint
-      run: golangci-lint run --timeout=2m --tests=false --enable=goconst,gofmt,goimports,golint,unconvert,unparam
+      run: |
+        golangci-lint run --timeout=2m --tests=false --enable=goconst,gofmt,goimports,golint,unconvert,unparam \
+          --exclude "Error return value of .((w|r)\.CloseWithError|viper.BindEnv). is not checked"
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -29,4 +29,4 @@ jobs:
       run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.24.0
 
     - name: Run golangci-lint
-      run: golangci-lint run --tests=false --enable=goconst,gofmt,goimports,golint,unconvert,unparam,bodyclose,gocyclo,misspell,maligned,whitespace,dupl,scopelint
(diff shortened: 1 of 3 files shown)


drakkan:
Thanks, I like the idea, I would like to add more linters too, anyway I need some time to check the current code before accepting the pull request. 

In the current code some warnings can be ignored, for example:

- `viper.BindEnv` can return an error only if we don't pass a key to bind
- `w.CloseWithError(err)` always return a nil error, and so on

enescakir:
You can list available linters with `golangci-lint linters`.

I exclude those two example. We can exclude more.
[Last run](https://github.com/enescakir/sftpgo/runs/629168668?check_suite_focus=true)

I can help to fix linter issues too.

You can move tests to Github Actions too. 
It can run tests with different Go versions (i.e. 1.12, 1.13, 1.14 etc.) in parallel.

drakkan:
Hi,

I fixed some of the lint issues with the latest commit, for other issues I disabled the check inside the code, for example:

```
go executeAction(operationAdd, user) //nolint:errcheck // the error is used in test cases only
```

in this case we can only log the error and this is already done inside the fuction itself. The returned error is used only inside test cases to check for wrong notification URL ecc.. 

The linter found some minor errors too, thanks!

So the code is now ready to accept this pull request, please remove the `exclude` added inside your last commit and change the command in this way:

```
golangci-lint run --tests=false --enable=goconst,gofmt,goimports,golint,unconvert,unparam,bodyclose,gocyclo,misspell,maligned,whitespace,dupl,scopelint
```

In future I would like to enable `golangci-lint` for test cases too but before I need to refactor them to use [testify](https://github.com/stretchr/testify)

enescakir:
Codebase is huge. Sometimes it could not lint in 1m. [Example](https://github.com/enescakir/sftpgo/pull/2/checks?check_run_id=633209390)

I increased it to 2m.

.github/workflows/quality.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Add GOPATH to PATH
+      run: echo "::add-path::$(go env GOPATH)/bin"
+      shell: bash
+
+    - name: Install golangci-lint
+      run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.24.0
```
drakkan, line 29: Do you have any reason to not use the latest golangcli-lint version?

drakkan:
thanks!

PR #113, opened 2020-04-28
fix log level changing problem

enescakir:
`Level()` creates new logger. It's not changing current one. Receiver isn't a pointer.
So changing log level isn't effective.

```golang
// Level creates a child logger with the minimum accepted level set to level.
func (l Logger) Level(lvl Level) Logger {
	l.level = lvl
	return l
}
```


drakkan:
Thanks!

PR #165, opened 2020-09-01
Allow disabling password authentication in sftpd

trapped:
Adds a `password_disabled` bool config key to `sftpd`, which causes `PasswordCallback` to be left `nil`, thus disabling password auth.

Useful in pubkey-only setups.

### DIFF
--- a/docs/full-configuration.md
+++ b/docs/full-configuration.md
@@ -63,6 +63,7 @@ The configuration file contains the following sections:
   - `bind_address`, string. Leave blank to listen on all available network interfaces. Default: ""
   - `idle_timeout`, integer. Deprecated, please use the same key in `common` section.
   - `max_auth_tries` integer. Maximum number of authentication attempts permitted per connection. If set to a negative number, the number of attempts is unlimited. If set to zero, the number of attempts are limited to 6.
+  - `password_disabled`, boolean. Set to false to forbid password authentication (for example in a pubkey-only setup).
   - `banner`, string. Identification string used by the server. Leave empty to use the default banner. Default `SFTPGo_<version>`, for example `SSH-2.0-SFTPGo_0.9.5`
   - `upload_mode` integer. Deprecated, please use the same key in `common` section.
   - `actions`, struct. Deprecated, please use the same key in `common` section.
--- a/sftpd/server.go
+++ b/sftpd/server.go
@@ -97,6 +97,8 @@ type Configuration struct {
 	// The following SSH commands are enabled by default: "md5sum", "sha1sum", "cd", "pwd".
 	// "*" enables all supported SSH commands.
 	EnabledSSHCommands []string `json:"enabled_ssh_commands" mapstructure:"enabled_ssh_commands"`
+	// PasswordDisabled specifies whether to forbid password authentication, for example in a publickey-only setup.
+	PasswordDisabled bool `json:"password_disabled" mapstructure:"password_disabled"`
 	// Absolute path to an external program or an HTTP URL to invoke for keyboard interactive authentication.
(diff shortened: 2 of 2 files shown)


drakkan:
Hi, thank your for your contribution, this global setting will disable multi step authentication too (for example public key + password). Are you aware that you can disable login methods per-user? Anyway if we want to add a global setting like this I would try to keep multi step authentication working

trapped:
@drakkan Hi, thanks for the quick reply!

Just to give a bit of context: I'm running `sftpgo` as an SFTP gateway to S3 and using external auth (API hook) exclusively. I'm aware that the hook is called per-user, and I'm currently using it for pubkey auth.

The issue I'm running into is that password authentication, even if in my case is never allowed, is still advertised on connection; logs from `sftp -v localhost:22`:

with `PasswordCallback = nil`:
```
debug1: Host 'localhost' is known and matches the RSA host key.
debug1: Found key in /home/user/.ssh/known_hosts:43
... debug messages ...
debug1: SSH2_MSG_SERVICE_ACCEPT received
Login banner!
debug1: Authentications that can continue: publickey
debug1: Next authentication method: publickey
debug1: Offering public key: RSA SHA256:xxx /home/user/.ssh/id_rsa
debug1: Authentications that can continue: publickey
debug1: Trying private key: /home/user/.ssh/id_dsa
debug1: Trying private key: /home/user/.ssh/id_ecdsa
debug1: Trying private key: /home/user/.ssh/id_ed25519
debug1: No more authentication methods to try.
user@localhost: Permission denied (publickey).
Connection closed
```

while with it set to the default callback handler it advertises password auth (and the OpenSSH client prompts for it by default):
```
debug1: Host 'localhost' is known and matches the RSA host key.
debug1: Found key in /home/user/.ssh/known_hosts:43
... debug messages ...
debug1: SSH2_MSG_SERVICE_ACCEPT received
Login banner!
debug1: Authentications that can continue: password,publickey
debug1: Next authentication method: publickey
debug1: Offering public key: RSA SHA256:xxx /home/user/.ssh/id_rsa
debug1: Authentications that can continue: password,publickey
debug1: Trying private key: /home/user/.ssh/id_dsa
debug1: Trying private key: /home/user/.ssh/id_ecdsa
debug1: Trying private key: /home/user/.ssh/id_ed25519
debug1: Next authentication method: password
user@localhost's password:
```

Normally this wouldn't be an issue as SSH clients can just set `-o PasswordAuthentication=no`, but some of them are running legacy code I've got no control over that apparently either fails or hangs if password auth is advertised, so I'm trying to work around that.
Setting `password_disabled` has the same effect as `PasswordAuthentication no` (in `/etc/ssh/sshd_config`) in this scenario.

I'm already using this setting as-is in a fork, but I'm totally open to improving this PR (for example for preserving multistep auth)! Let me know how you would go about it.

Grazie!

drakkan:
Hi,

I mean that you can add a per-user denied login methods, but this way password login method will still be advertised but refused, so it will not solve your issue. Please let me some time to understand if there is a better way to handle this, if we disable password callback multi step auth will not be possibile anymore this should be at least documented

drakkan:
Hi, 

crypto/ssh will always advertise password login method if password callback is not nil. I'll merge this pull request as is and I'll push a new commit with the following changes:

- rename the setting to PasswordAuthentication as OpenSSH
- avoid to advertise password login method in multi-step auth if it is disabled
- add some test cases

Please let me know if it works with your clients. I'm unable to reproduce your issue, thank you

trapped:
@drakkan just tried that commit and it seems to work great. Thanks!

PR #174, opened 2020-09-21
Set verbosity for go commands in docker build

smueller18:
To visualize the progress of the docker builds, I added the `-v` flags to the go commands.


drakkan:
Thank you!

Issue #178, opened 2020-09-28
GCS Wrong Content-Type


thomas-advantitge:
Some file types have a non-standard Content-Type after upload through the SFTP server:
![image](https://user-images.githubusercontent.com/66438062/94419469-1afc0680-0183-11eb-9299-b7367538c219.png)
It also seems to differ for the same file type:
![image](https://user-images.githubusercontent.com/66438062/94419727-762df900-0183-11eb-8d4d-2283369c4f4d.png)

Wrong content-type values causes browsers like Chrome to download the file with the extension related to the Content-Type (e.g. the application/zip content-type file translates to a file with extension ".zip", causing troubles opening the file).

I took a look in gcsfs.go but didn't see any content-type related setters for the Create method, so not sure the content-type "guessing" is related to sftpgo or the golang client library for Google Cloud Storage.

drakkan:
Hi, 

SFTPGo does not set a content type after creating an [object writer](https://github.com/drakkan/sftpgo/blob/master/vfs/gcsfs.go#L183). It only set `StorageClass` if appropriate.

Based on the doc [here](https://github.com/googleapis/google-cloud-go/blob/master/storage/storage.go#L1004)

```
If no ContentType attribute is specified, the content type will be automatically sniffed using net/http.DetectContentType.
```

`DetectContentType` documentation is [here](https://golang.org/pkg/net/http/#DetectContentType). I don't know if I can fix this on the SFTPGo side: I can explicitly set a content type but how can I detect it? If I use `net/http.DetectContentType` we'll get the same behaviour.

I'll try to better investigate better later, maybe I can try to guess the mime type by [extension](https://github.com/drakkan/sftpgo/blob/master/webdavd/file.go#L67), are you able to do some tests using this method? 

Here is an untested patch

```
diff --git a/vfs/gcsfs.go b/vfs/gcsfs.go
index 7ac29e8..dc43799 100644
--- a/vfs/gcsfs.go
+++ b/vfs/gcsfs.go
@@ -7,6 +7,7 @@ import (
        "errors"
        "fmt"
        "io"
+       "mime"
        "net/http"
        "os"
        "path"
@@ -181,6 +182,10 @@ func (fs GCSFs) Create(name string, flag int) (*os.File, *PipeWriter, func(), er
        obj := bkt.Object(name)
        ctx, cancelFn := context.WithCancel(context.Background())
        objectWriter := obj.NewWriter(ctx)
[...]

Issue #124, opened 2020-05-28
Host keys auto generator outside config dir


sergiz:
https://github.com/drakkan/sftpgo/blob/a08dd85efd2b65bbe170b116ae42230a651143da/sftpd/server.go#L518-L536

Looking at this, it seems that you can utilize auto generation of the host keys, only if host_keys setting is empty. Otherwise you have to provide your own.

Would it be possible to add another setting, so we can specify a directory where our host keys should be regardless of our config dir, and if they don't exist they are automatically created?

drakkan:
> https://github.com/drakkan/sftpgo/blob/a08dd85efd2b65bbe170b116ae42230a651143da/sftpd/server.go#L518-L536
> 
> Looking at this, it seems that you can utilize auto generation of the host keys, only if host_keys setting is empty. Otherwise you have to provide your own.
> 

Correct. If you use the default configuration then an RSA/2048 and a ECDSA/P-256 key will be automatically generated inside the config dir (this is a sane default very similar to OpenSSH), otherwise if you specify a relative or absolute path the keys must be there.

> Would it be possible to add another setting, so we can specify a directory where our host keys should be regardless of our config dir, and if they don't exist they are automatically created?

If you specify a path for your keys you probably want a specific key size/algorithm (different from the default ones), I don't want to expose all these details or try to deduce them from the key name. If you specify only one key it must be RSA (2048? 4096?...) or EDSA (P-256?,P-384?,P-512?...,) and if you specify 3 keys?

Eventually I could add a sub command to generate keys but  could this be better than `ssh-keygen`?

sergiz:
I'm not asking for any advanced thing, a simple string setting to specify a path where keys should be.
Basically just this
`autoFile := filepath.Join(configDir, k)` -> if host_keys_path (or something alike) not null, then use it, otherwise use configDir.

Don't need to touch anything else. If you need more advanced configuration you have host_keys setting already.

drakkan:
Hi,

If you want your keys auto created inside inside another directory you can use an hack such as this one, from the config directory create these symlinks to missing files:

```
ln -s /tmp/id_rsa . 
ln -s /tmp/id_rsa.pub .
ln -s /tmp/id_ecdsa .
ln -s /tmp/id_ecdsa.pub .
```

and start sftpgo, the keys will be created in /tmp.

[...]

PR #180, opened 2020-10-04
Add Dockerfile and GitHub Actions workflow

sagikazarmark:
This PR adds a Dockerfile to the project. It's largely based on my previous Go applications and the existing examples in the repo. It's not perfect though, so we might need some fine tuning in the future, particularly around volumes and configuration.

Open questions:

- What are backups? Are they supposed to be saved to persistent storage or are they only meaningful during the lifecycle of the application?
- Where does the credential path point? What's being read from there? (Existing dockerfiles contain no reference)

Currently the above settings can only be overridden from environment variables, because defaults are defined for them in env vars. (I think that's an acceptable compromise though).

I'll work with this image anyway, so I'll provide patches along the way.

The PR also adds a GitHub Actions workflow for building Docker images based on the official example (I removed a few parts):

https://github.com/docker/build-push-action#complete-workflow

After merging the PR, @drakkan , you'll need to do some manual steps though:

1. Login to ghcr.io: `docker login ghcr.io`. You'll need a GH PAT for that (with package write scope)
1. Build the image locally: `docker build -t ghcr.io/drakkan/sftpgo:edge .`
1. Push the image `docker push ghcr.io/drakkan/sftpgo:edge`
1. Go to packages under your [account](https://github.com/drakkan?tab=packages)
1. Make the package public
1. Configure a secret called `CR_PAT` with package write scope under [secrets](https://github.com/drakkan/sftpgo/settings/secrets)
1. Uncomment the relevant parts in docker.yml (login and push) (happy to review a PR)

These steps are necessary, because GitHub packages seem to be super private for the first time, so you need to push and publish manually. (Although rereading the above: you should also be fine to do step 6 and 7 first, since `CR_PAT` will be your token anyway, and then step 5 to publish the package. I had some trouble with packages under organizations, but this is under your personal account, so it should be fine)

Sorry about the complicated process, let me know if you need any help.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,9 +12,13 @@ ARG GOPROXY
 COPY go.mod go.sum ./
 RUN go mod download
 
+ARG COMMIT_SHA
+
 COPY . .
 
-RUN go build -ldflags "-s -w -X github.com/drakkan/sftpgo/version.commit=`git describe --always --dirty` -X github.com/drakkan/sftpgo/version.date=`date -u +%FT%TZ`" -o sftpgo
+RUN set -xe && \
+    export COMMIT_SHA=${COMMIT_SHA:-$(git describe --always --dirty)} && \
+    go build -ldflags "-s -w -X github.com/drakkan/sftpgo/version.commit=${COMMIT_SHA} -X github.com/drakkan/sftpgo/version.date=`date -u +%FT%TZ`" -o sftpgo
 
 
 FROM alpine:3.12
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,7 +48,6 @@ ENV SFTPGO_HTTPD__STATIC_FILES_PATH=/srv/sftpgo/web/static
 # Sane defaults, but users should still be able to override this from env vars
 ENV SFTPGO_DATA_PROVIDER__USERS_BASE_DIR=/var/lib/sftpgo/data
 ENV SFTPGO_DATA_PROVIDER__CREDENTIALS_PATH=/var/lib/sftpgo/credentials
-ENV SFTPGO_DATA_PROVIDER__NAME=/var/lib/sftpgo/sftpgo.db
 ENV SFTPGO_HTTPD__BACKUPS_PATH=/var/lib/sftpgo/backups
 ENV SFTPGO_SFTPD__HOST_KEYS=/var/lib/sftpgo/host_keys/id_rsa,/var/lib/sftpgo/host_keys/id_ecdsa
 
@@ -56,6 +55,8 @@ COPY --from=builder /workspace/sftpgo.json /etc/sftpgo/sftpgo.json
 COPY --from=builder /workspace/templates /srv/sftpgo/web/templates
 COPY --from=builder /workspace/static /srv/sftpgo/web/static
 COPY --from=builder /workspace/sftpgo /usr/local/bin/
+
+RUN sed -i "s|sftpgo.db|/var/lib/sftpgo/sftpgo.db|" /etc/sftpgo/sftpgo.json
 
 RUN chown -R sftpgo:sftpgo /etc/sftpgo /var/lib/sftpgo /srv/sftpgo/web
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,8 +55,6 @@ COPY --from=builder /workspace/templates /srv/sftpgo/web/templates
 COPY --from=builder /workspace/static /srv/sftpgo/web/static
 COPY --from=builder /workspace/sftpgo /usr/local/bin/
 
-RUN sed -i "s/sqlite/bolt/; s|sftpgo.db|/var/lib/sftpgo/sftpgo.db|" /etc/sftpgo/sftpgo.json
-
 RUN chown -R sftpgo:sftpgo /etc/sftpgo /var/lib/sftpgo /srv/sftpgo/web
 
 USER sftpgo
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ SHELL ["/bin/bash", "-c"]
 # https://github.com/gliderlabs/docker-alpine/issues/367#issuecomment-424546457
 RUN test ! -e /etc/nsswitch.conf && echo 'hosts: files dns' > /etc/nsswitch.conf
 
-RUN mkdir -p /etc/sftpgo /var/lib/sftpgo/{backups,data,host_keys} /srv/sftpgo/web
+RUN mkdir -p /etc/sftpgo /var/lib/sftpgo/{backups,data,credentials,host_keys} /srv/sftpgo/web
 
 RUN addgroup -g 1000 -S sftpgo
 RUN adduser -u 1000 -h /var/lib/sftpgo -s /sbin/nologin -G sftpgo -S -D -H sftpgo
@@ -47,6 +47,7 @@ ENV SFTPGO_HTTPD__STATIC_FILES_PATH=/srv/sftpgo/web/static
 
 # Sane defaults, but users should still be able to override this from env vars
 ENV SFTPGO_DATA_PROVIDER__USERS_BASE_DIR=/var/lib/sftpgo/data
+ENV SFTPGO_DATA_PROVIDER__CREDENTIALS_PATH=/var/lib/sftpgo/credentials
 ENV SFTPGO_DATA_PROVIDER__NAME=/var/lib/sftpgo/sftpgo.db
 ENV SFTPGO_HTTPD__BACKUPS_PATH=/var/lib/sftpgo/backups
 ENV SFTPGO_SFTPD__HOST_KEYS=/var/lib/sftpgo/host_keys/id_rsa,/var/lib/sftpgo/host_keys/id_ecdsa
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ SHELL ["/bin/bash", "-c"]
 # https://github.com/gliderlabs/docker-alpine/issues/367#issuecomment-424546457
 RUN test ! -e /etc/nsswitch.conf && echo 'hosts: files dns' > /etc/nsswitch.conf
 
-RUN mkdir -p /etc/sftpgo /var/lib/sftpgo/{backups,data} /srv/sftpgo/web
+RUN mkdir -p /etc/sftpgo /var/lib/sftpgo/{backups,data,host_keys} /srv/sftpgo/web
 
 RUN addgroup -g 1000 -S sftpgo
 RUN adduser -u 1000 -h /srv/sftpgo -s /sbin/nologin -G sftpgo -S -D -H sftpgo
@@ -47,6 +47,7 @@ ENV SFTPGO_HTTPD__STATIC_FILES_PATH=/srv/sftpgo/web/static
 # Sane defaults, but users should still be able to override this from env vars
 ENV SFTPGO_DATA_PROVIDER__USERS_BASE_DIR=/var/lib/sftpgo/data
 ENV SFTPGO_HTTPD__BACKUPS_PATH=/var/lib/sftpgo/backups
+ENV SFTPGO_SFTPD__HOST_KEYS=/var/lib/sftpgo/host_keys/id_rsa,/var/lib/sftpgo/host_keys/id_ecdsa
 
 COPY --from=builder /workspace/sftpgo.json /etc/sftpgo/sftpgo.json
 COPY --from=builder /workspace/templates /srv/sftpgo/web/templates
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,6 +52,8 @@ COPY --from=builder /workspace/templates /srv/sftpgo/web/templates
 COPY --from=builder /workspace/static /srv/sftpgo/web/static
 COPY --from=builder /workspace/sftpgo /usr/local/bin/
 
+RUN sed -i "s/sqlite/bolt/; s|sftpgo.db|/var/lib/sftpgo/sftpgo.db|" /etc/sftpgo/sftpgo.json
+
 RUN chown -R sftpgo:sftpgo /etc/sftpgo /var/lib/sftpgo /srv/sftpgo/web
 
 USER sftpgo
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,12 +31,12 @@ SHELL ["/bin/bash", "-c"]
 # https://github.com/gliderlabs/docker-alpine/issues/367#issuecomment-424546457
 RUN test ! -e /etc/nsswitch.conf && echo 'hosts: files dns' > /etc/nsswitch.conf
 
-RUN mkdir -p /data /etc/sftpgo /srv/sftpgo/web /srv/sftpgo/backups
+RUN mkdir -p /data /etc/sftpgo /var/lib/sftpgo/backups /srv/sftpgo/web
 
 RUN addgroup -g 1000 -S sftpgo
 RUN adduser -u 1000 -h /srv/sftpgo -s /sbin/nologin -G sftpgo -S -D -H sftpgo
 
-VOLUME ["/data", "/srv/sftpgo/backups"]
+VOLUME ["/data", "/var/lib/sftpgo"]
 
 # Override some configuration details
 ENV SFTPGO_CONFIG_DIR=/etc/sftpgo
@@ -46,14 +46,14 @@ ENV SFTPGO_HTTPD__STATIC_FILES_PATH=/srv/sftpgo/web/static
 
 # Sane defaults, but users should still be able to override this from env vars
 ENV SFTPGO_DATA_PROVIDER__USERS_BASE_DIR=/data
-ENV SFTPGO_HTTPD__BACKUPS_PATH=/srv/sftpgo/backups
+ENV SFTPGO_HTTPD__BACKUPS_PATH=/var/lib/sftpgo/backups
 
 COPY --from=builder /workspace/sftpgo.json /etc/sftpgo/sftpgo.json
 COPY --from=builder /workspace/templates /srv/sftpgo/web/templates
 COPY --from=builder /workspace/static /srv/sftpgo/web/static
 COPY --from=builder /workspace/sftpgo /usr/local/bin/
 
-RUN chown -R sftpgo:sftpgo /data /etc/sftpgo /srv/sftpgo/web /srv/sftpgo/backups
+RUN chown -R sftpgo:sftpgo /data /etc/sftpgo /var/lib/sftpgo /srv/sftpgo/web
 
 USER sftpgo
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,56 @@
+FROM golang:1.15-alpine AS builder
+
+ENV GOFLAGS="-mod=readonly"
+
+RUN apk add --update --no-cache bash ca-certificates curl git gcc g++
+
+RUN mkdir -p /workspace
+WORKDIR /workspace
+
+ARG GOPROXY
+
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+
+RUN go build -ldflags "-s -w -X github.com/drakkan/sftpgo/version.commit=`git describe --always --dirty` -X github.com/drakkan/sftpgo/version.date=`date -u +%FT%TZ`" -o sftpgo
+
+
+FROM alpine:3.12
+
+RUN apk add --update --no-cache ca-certificates tzdata bash
+
+SHELL ["/bin/bash", "-c"]
+
+# set up nsswitch.conf for Go's "netgo" implementation
+# https://github.com/gliderlabs/docker-alpine/issues/367#issuecomment-424546457
+RUN test ! -e /etc/nsswitch.conf && echo 'hosts: files dns' > /etc/nsswitch.conf
+
+RUN mkdir -p /data /etc/sftpgo /srv/sftpgo/web /srv/sftpgo/backups
+
+RUN addgroup -g 1000 -S sftpgo
+RUN adduser -u 1000 -h /srv/sftpgo -s /sbin/nologin -G sftpgo -S -D -H sftpgo
+
+VOLUME ["/data", "/srv/sftpgo/backups"]
+
+# Override some configuration details
+ENV SFTPGO_CONFIG_DIR=/etc/sftpgo
+ENV SFTPGO_LOG_FILE_PATH=""
+ENV SFTPGO_HTTPD__TEMPLATES_PATH=/srv/sftpgo/web/templates
+ENV SFTPGO_HTTPD__STATIC_FILES_PATH=/srv/sftpgo/web/static
+
+# Sane defaults, but users should still be able to override this from env vars
+ENV SFTPGO_DATA_PROVIDER__USERS_BASE_DIR=/data
+ENV SFTPGO_HTTPD__BACKUPS_PATH=/srv/sftpgo/backups
+
+COPY --from=builder /workspace/sftpgo.json /etc/sftpgo/sftpgo.json
+COPY --from=builder /workspace/templates /srv/sftpgo/web/templates
+COPY --from=builder /workspace/static /srv/sftpgo/web/static
+COPY --from=builder /workspace/sftpgo /usr/local/bin/
+
+RUN chown -R sftpgo:sftpgo /data /etc/sftpgo /srv/sftpgo/web /srv/sftpgo/backups
+
+USER sftpgo
+
+CMD sftpgo serve
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,7 @@ ENV SFTPGO_HTTPD__STATIC_FILES_PATH=/srv/sftpgo/web/static
 
 # Sane defaults, but users should still be able to override this from env vars
 ENV SFTPGO_DATA_PROVIDER__USERS_BASE_DIR=/var/lib/sftpgo/data
+ENV SFTPGO_DATA_PROVIDER__NAME=/var/lib/sftpgo/sftpgo.db
 ENV SFTPGO_HTTPD__BACKUPS_PATH=/var/lib/sftpgo/backups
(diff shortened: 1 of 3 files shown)


drakkan:
Hi, 

thank you for your PR.

While trying to build it locally using this command:

```
docker build -t ghcr.io/drakkan/sftpgo:edge .
```

I get this error:

```
Step 6/32 : ARG GOPROXY
 ---> Running in f8d5a9be3075
Removing intermediate container f8d5a9be3075
 ---> b340ef6bfd56
Step 7/32 : COPY go.mod go.sum ./
COPY failed: stat /var/lib/docker/tmp/docker-builder424797755/go.mod: no such file or directory
```

drakkan:
> This PR adds a Dockerfile to the project. It's largely based on my previous Go applications and the existing examples in the repo. It's not perfect though, so we might need some fine tuning in the future, particularly around volumes and configuration.
> 
> Open questions:
> 
> * What are backups? Are they supposed to be saved to persistent storage or are they only meaningful during the lifecycle of the application?

backups are supposed to be saved to a persistent storage. You can backup your users and folders and the resulting JSON file will be saved inside the backup path

> * Where does the credential path point? What's being read from there? (Existing dockerfiles contain no reference)
> 

the credential path currently store GCS credential files.

In standard Linux packages both backups and credentials are stored inside /var/lib/sftpgo maybe we should export this path as volume

> Currently the above settings can only be overridden from environment variables, because defaults are defined for them in env vars. (I think that's an acceptable compromise though).
> 
> I'll work with this image anyway, so I'll provide patches along the way.
> 
> The PR also adds a GitHub Actions workflow for building Docker images based on the official example (I removed a few parts):
> 
> https://github.com/docker/build-push-action#complete-workflow
> 
> After merging the PR, @drakkan , you'll need to do some manual steps though:
> 
> 1. Login to ghcr.io: `docker login ghcr.io`. You'll need a GH PAT for that (with package write scope)
> 2. Build the image locally: `docker build -t ghcr.io/drakkan/sftpgo:edge .`
> 3. Push the image `docker push ghcr.io/drakkan/sftpgo:edge`
> 4. Go to packages under your [account](https://github.com/drakkan?tab=packages)
> 5. Make the package public
> 6. Configure a secret called `CR_PAT` with package write scope under [secrets](https://github.com/drakkan/sftpgo/settings/secrets)
> 7. Uncomment the relevant parts in docker.yml (login and push) (happy to review a PR)
> 
> These steps are necessary, because GitHub packages seem to be super private for the first time, so you need to push and publish manually. (Although rereading the above: you should also be fine to do step 6 and 7 first, since `CR_PAT` will be your token anyway, and then step 5 to publish the package. I had some trouble with packages under organizations, but this is under your personal account, so it should be fine)
> 
> Sorry about the complicated process, let me know if you need any help.

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+ARG COMMIT_SHA
+
+COPY . .
+
+RUN set -xe && \
+    export COMMIT_SHA=${COMMIT_SHA:-$(git describe --always --dirty)} && \
+    go build -ldflags "-s -w -X github.com/drakkan/sftpgo/version.commit=${COMMIT_SHA} -X github.com/drakkan/sftpgo/version.date=`date -u +%FT%TZ`" -o sftpgo
+
+
+FROM alpine:3.12
```
drakkan, line 24: as above I would prefer alpine:latest

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+RUN adduser -u 1000 -h /srv/sftpgo -s /sbin/nologin -G sftpgo -S -D -H sftpgo
+
+VOLUME ["/data", "/srv/sftpgo/backups"]
+
+# Override some configuration details
+ENV SFTPGO_CONFIG_DIR=/etc/sftpgo
+ENV SFTPGO_LOG_FILE_PATH=""
+ENV SFTPGO_HTTPD__TEMPLATES_PATH=/srv/sftpgo/web/templates
+ENV SFTPGO_HTTPD__STATIC_FILES_PATH=/srv/sftpgo/web/static
+
```
drakkan, line 46: templates and static files are not supposed to be modified, so we don't need persistent storage for them.

We need persistent storage for SSH host keys, we can set an env var such as this one:

```
SFTPGO_SFTPD__HOST_KEYS=/persistent/path/id_rsa,/persistent/path/id_ecdsa
```

to generate/read them from a persistent path

drakkan (changes requested):

sagikazarmark:
That error is really weird. Are you trying to run it from the root of the project? The Dockerfile doesn't download the source code (compared to the existing examples).

As for using latest versions: I have really bad experience with that. There was almost no new minor alpine version when the build just didn't magically break. Same for the Go version: I almost always tweaked something when a new release came out.

Pinning versions provide reproducible(ish) builds, so I'd recommend pinning at least the minor version in both cases (I'd even use `golang:1.15-alpine3.12` for the Go image), but I'll change it if you would like me to.

As for persistent storage: we have a bunch of different options here. Since the process in the container actually needs to write files, permissions complicate things.

One option is to set a preferred location for all persistent storage paths (set env vars, so config files can't override them) and require users to mount their volumes there. This requires a little bit more documentation and care as users tend to mount their volumes at some random path and configure it manually (which will only work if they set env vars).

Alternatively, we could prepare the default config file with those values (so they are not env vars => users can override config from file, but then they always have to specify those paths which is less desirable IMHO)

> the credential path currently store GCS credential files.

But they are only being read from here, right? Nothing is written here. In that case, users will have mount their credentials here anyway, so I wouldn't worry about it as long as sftpgo doesn't crash if there is nothing there. :) (in fact, I really hope secrets aren't supposed to be written to persistent storage 😄 )

As for host keys, I know sftpgo can generate its own keys, but I guess the recommendation here (in a production environment) is to generate your own and mount in the container. When you mount the keys, you don't actually need persistent storage (eg. mount secrets in kubernetes). So I would find a default path for host keys, make that path a volume and then recommend users to mount their own.

Let me know what you think! I'll try to improve the persistent path settings.

The overall decision we need to make is whether we are comfortable with setting some default stuff in env vars and require users to set those keys using env vars only (since env vars override config file values). (I don't see any alternative at the moment)

drakkan:
> That error is really weird. Are you trying to run it from the root of the project? The Dockerfile doesn't download the source code (compared to the existing examples).
> 

No sorry, I tryed running outside the source tree as with the previous Dockerfiles. Running from the source tree works as expected. Sorry for the noise.

> As for using latest versions: I have really bad experience with that. There was almost no new minor alpine version when the build just didn't magically break. Same for the Go version: I almost always tweaked something when a new release came out.
> 
> Pinning versions provide reproducible(ish) builds, so I'd recommend pinning at least the minor version in both cases (I'd even use `golang:1.15-alpine3.12` for the Go image), but I'll change it if you would like me to.
> 

OK, I trust your expierence.

> As for persistent storage: we have a bunch of different options here. Since the process in the container actually needs to write files, permissions complicate things.
> 
> One option is to set a preferred location for all persistent storage paths (set env vars, so config files can't override them) and require users to mount their volumes there. This requires a little bit more documentation and care as users tend to mount their volumes at some random path and configure it manually (which will only work if they set env vars).
> 

we could keep the default config file inside the container, so users can still override it if they want, and use the same strategy that I use for distro packages:

- all the variable data are inside `/var/lib/sftpgo`, for example:

```
# cd /var/lib/sftpgo
# ls -la
totale 60
drwxr-x---  6 sftpgo sftpgo  4096 22 set 15.17 .
drwxr-xr-x 33 root   root    4096  5 ott 09.06 ..
drwxr-xr-x  2 sftpgo sftpgo  4096 22 set 15.20 .aws
drwx------  2 sftpgo sftpgo  4096 22 set 15.36 backups
drwx------  2 sftpgo sftpgo  4096 19 set 10.27 credentials
-rw-r--r--  1 sftpgo sftpgo 65536 25 set 18.27 sftpgo.db
drwxr-x---  3 sftpgo sftpgo  4096 19 set 10.33 users
```

where `.aws` stores aws credentials, `sftpgo.db` is the SQLite database, `users` is the base dir for all the local users, etc..

This way we can export only one volume, what do you think about?

> Alternatively, we could prepare the default config file with those values (so they are not env vars => users can override config from file, but then they always have to specify those paths which is less desirable IMHO)
> 
> > the credential path currently store GCS credential files.
> 
> But they are only being read from here, right? Nothing is written here. In that case, users will have mount their credentials here anyway, so I wouldn't worry about it as long as sftpgo doesn't crash if there is nothing there. :) (in fact, I really hope secrets aren't supposed to be written to persistent storage  )

Credentials are written when you create a new GCS user (if a credential file is provided) and are read when the GCS user logins.

> 
> As for host keys, I know sftpgo can generate its own keys, but I guess the recommendation here (in a production environment) is to generate your own and mount in the container. When you mount the keys, you don't actually need persistent storage (eg. mount secrets in kubernetes). So I would find a default path for host keys, make that path a volume and then recommend users to mount their own.
> 

this could be acceptable, see also #124 for a different point of view but probably host keys can be mounted as explained [here](https://github.com/atmoz/sftp#providing-your-own-ssh-host-key-recommended)

> Let me know what you think! I'll try to improve the persistent path settings.
> 
> The overall decision we need to make is whether we are comfortable with setting some default stuff in env vars and require users to set those keys using env vars only (since env vars override config file values). (I don't see any alternative at the moment)

drakkan:
some other notes:

- we could add `mailcap` package to the default image so we have /etc/mime.types that is useful to guess mime types, see #178 . It is used for WebDAV too
- we could add `rsync` so it is available if needed
- we could add `git` so the image can be used to host git repos too

I could add these packages after merging the PR if you prefer, thank you!

sagikazarmark:
> we could keep the default config file inside the container, so users can still override it if they want, and use the same strategy that I use for distro packages

My thoughts exactly. I reorganized the directory structure in the image.

Some additional changes:
- I installed the requested packages
- switched to boltdb as a default driver to avoid the need for initialization (saved in `/var/lib/sftpgo`)

To be honest, the credential bit worries me a little: under no circumstances should be secrets written to persistent storage.

At least not by default. So how about the keep the home directory of the `sftpgo` user in `/home` (currently it's `/var/lib/sftpgo` and set the credentials path to point there.

(To be honest, it's still a bit blurry how it works. For example if I use the instance metadata of a GCP instance to authenticate against GCP, are there any credential files written there?)

drakkan:
I think [this](https://github.com/drakkan/sftpgo/commit/3e2afc35ba168a9dd283b2bd4785e6f8a8a12e3f) change could help here

sagikazarmark:
Oh, I'll most definitely want to disable that in production. 😄 So please add a flag for it, I'm not comfortable with an application trying to run migrations in a production environment automatically.

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+ENV SFTPGO_DATA_PROVIDER__USERS_BASE_DIR=/var/lib/sftpgo/data
+ENV SFTPGO_HTTPD__BACKUPS_PATH=/var/lib/sftpgo/backups
+ENV SFTPGO_SFTPD__HOST_KEYS=/var/lib/sftpgo/host_keys/id_rsa,/var/lib/sftpgo/host_keys/id_ecdsa
+
+COPY --from=builder /workspace/sftpgo.json /etc/sftpgo/sftpgo.json
+COPY --from=builder /workspace/templates /srv/sftpgo/web/templates
+COPY --from=builder /workspace/static /srv/sftpgo/web/static
+COPY --from=builder /workspace/sftpgo /usr/local/bin/
+
+RUN sed -i "s/sqlite/bolt/; s|sftpgo.db|/var/lib/sftpgo/sftpgo.db|" /etc/sftpgo/sftpgo.json
```
drakkan, line 58: data provider can be changed using env vars, sed is not required, anyway after my latest change we should be able to use sqlite provider too, manual initialization is not required anymore

sagikazarmark: I'd stick to modifying the default config wherever possible or when the user is likely to override that anyway, hence I used sed. In a preview environment it doesn't really matter if it's sqlite or bolt (although bolt kinda sounds better for me, but not for any apparent reason), but in production I'm not sure I'd want to let an app modify the database automatically.

drakkan:
> Oh, I'll most definitely want to disable that in production.  So please add a flag for it, I'm not comfortable with an application trying to run migrations in a production environment automatically.

Hi, in the first versions SFTPGo users had to manually initialize the database and run the database migrations (there were SQL files for each supported data provider inside the repo). 

Before the latest change the database was manually initialized using the `initprovider` command to version `1` and then automatically migrated on startup. Now I have tryed to avoid the initial manual step too. If you upgrade SFTPGo from version 0.9.6 to 1.0.0 for example the database is automatically migrated. The automatic migration happen for bolt provider too, see [here](https://github.com/drakkan/sftpgo/blob/master/dataprovider/bolt.go#L709).

I'll evaluate to add an CLI flag but it would be disabled by default since most users want to avoid manual configuration as much as possible

sagikazarmark:
> I'll evaluate to add an CLI flag but it would be disabled by default since most users want to avoid manual configuration as much as possible

As long as I have control over it, I don't mind if I have to manually disable it.

drakkan:
Hi, I did a quick test using the provided Dockerfile, there are some minor things that I can fix after merging myself, I think the only missing things are the following ones:

1) sqlite/bolt database should be moved to persistent storage
2) some docs to explain how to use/customize it

thank you!

sagikazarmark:
> sqlite/bolt database should be moved to persistent storage

They are, `/var/lib/sftpgo` should be considered as persistent store

> To be honest, the credential bit worries me a little: under no circumstances should be secrets written to persistent storage.
> 
> At least not by default. So how about the keep the home directory of the `sftpgo` user in `/home` (currently it's `/var/lib/sftpgo` and set the credentials path to point there.
> 
> (To be honest, it's still a bit blurry how it works. For example if I use the instance metadata of a GCP instance to authenticate against GCP, are there any credential files written there?)

What do you think about this? ☝️ That's my last blocker for this PR.

> some docs to explain how to use/customize it

That's fair, but I'd delay it to a different PR.

sagikazarmark:
I was wondering if the default config should be added by default. As far as I can tell, sftpgo sets defaults internally, so how about adding a config file like this:

```json
{
  "data_provider": {
    "driver": "sqlite",
    "name": "/var/lib/sftpgo/sftpgo.db"
  }
}
```

The data provider is likely to be changed by the user.

drakkan:
> > sqlite/bolt database should be moved to persistent storage
> 
> ~They are, `/var/lib/sftpgo` should be considered as persistent store~
> I just realized I reverted that change. I'll update it in a moment
> 
> > To be honest, the credential bit worries me a little: under no circumstances should be secrets written to persistent storage.
> > At least not by default. So how about the keep the home directory of the `sftpgo` user in `/home` (currently it's `/var/lib/sftpgo` and set the credentials path to point there.
> > (To be honest, it's still a bit blurry how it works. For example if I use the instance metadata of a GCP instance to authenticate against GCP, are there any credential files written there?)
> 
> What do you think about this?  That's my last blocker for this PR.

please take a look [here](https://github.com/drakkan/sftpgo/blob/master/vfs/gcsfs.go#L61) we can use automatic credentials or file credentials. If you configure a GCS user using file credentials then the file will be saved inside the `credentials` dir. This file is required when the user logins so it must be persistent. If you choose automatic credentials SFTPGo does not store nothing.

More detail [here](https://cloud.google.com/docs/authentication/production#providing_credentials_to_your_application)

> 
> > some docs to explain how to use/customize it
> 
> That's fair, but I'd delay it to a different PR.

OK

drakkan:
> I was wondering if the default config should be added by default. As far as I can tell, sftpgo sets defaults internally, so how about adding a config file like this:
> 
> ```json
> {
>   "data_provider": {
>     "driver": "sqlite",
>     "name": "/var/lib/sftpgo/sftpgo.db"
>   }
> }
> ```
> 

this is problematic for Windows users

> The data provider is likely to be changed by the user.
> 
> I'm afraid that setting these defaults as env vars will cause confusion for users: anything set as an env var can only be overridden by env vars. For some config (like paths) I guess it's ok, bot for other keys, likely to be changed by users, it could be confusing.

I have not yet tested this with this Dockerfile but if we ship `/etc/sftpgo/sftpgo.json` inside the image this will be the default config file (the users will not see it since /etc/sftpgo should not be mounted as volume) and all configurations can be set using env vars. The confusion should happen only if the users will open `/etc/sftpgo/sftpgo.json` inside the non-persistent storage. Am I missing something?

sagikazarmark:
Okay, I think I finally understood. :) I needed to see the UI to understand that it actually uploads a credential file.

You are right, that needs to be persistent, although I would argue that it's not really safe from a security point of view and disabling it would be a nice feature. (I might also contribute Vault integration at some point)

We should probably add a warning about this in the documentation.

> this is problematic for Windows users

I doubt this docker image will run on Windows

> the users will not see it since /etc/sftpgo should not be mounted as volume

Actually, users can still mount it with their config file. Writing a config file is often more readable for users than setting env vars. So if they do and they don't realise some configuration key have values in form of env vars inside the image, it might not be immediately clear why. Documentation can help though.

I guess we can iterate later as well.

drakkan:
> Okay, I think I finally understood. :) I needed to see the UI to understand that it actually uploads a credential file.
> 
> You are right, that needs to be persistent, although I would argue that it's not really safe from a security point of view and disabling it would be a nice feature. (I might also contribute Vault integration at some point)
> 
> We should probably add a warning about this in the documentation.
> 

Yes, we should advise that automatic configuration should be preferred. Optional Vault support would be great too!

> > this is problematic for Windows users
> 
> I doubt this docker image will run on Windows

ah ok, if you want to sed the default config file this is fine for me. For the Arch package I do exactly this:

https://aur.archlinux.org/cgit/aur.git/tree/sftpgo.json?h=sftpgo-git#n72

I would like to change the templates and static paths as for Arch PKGBUILD, but I could do this myself after merging.
Also creating directories inside /var/lib/sftpgo in the image should be useless if the user mount /var/lib/sftpgo.

uid/gid 1000 should be fine most of the time, env vars to allow to customize uid/gid could be useful too


> 
> > the users will not see it since /etc/sftpgo should not be mounted as volume
> 
> Actually, users can still mount it with their config file. Writing a config file is often more readable for users than setting env vars. So if they do and they don't realise some configuration key have values in form of env vars inside the image, it might not be immediately clear why. Documentation can help though.
> 
> I guess we can iterate later as well.

Yes, I agree

sagikazarmark:
> ah ok, if you want to sed the default config file this is fine for me. For the Arch package I do exactly this:

OK, updated using sed.

>  Also creating directories inside /var/lib/sftpgo in the image should be useless if the user mount /var/lib/sftpgo.

Right, but it's explicit and works OOTB if you don't mount anything.

> uid/gid 1000 should be fine most of the time, env vars to allow to customize uid/gid could be useful too

That would complicate things, it would require an entrypoint, we might not even be able to use the current `USER` approach.
Can we do it in a later PR if it turns out to be a requirement for someone? (I actually doubt it's useful to anyone other than a local development environment)

drakkan:
> > ah ok, if you want to sed the default config file this is fine for me. For the Arch package I do exactly this:
> 
> OK, updated using sed.
> 
> > Also creating directories inside /var/lib/sftpgo in the image should be useless if the user mount /var/lib/sftpgo.
> 
> Right, but it's explicit and works OOTB if you don't mount anything.
> 
> > uid/gid 1000 should be fine most of the time, env vars to allow to customize uid/gid could be useful too
> 
> That would complicate things, it would require an entrypoint, we might not even be able to use the current `USER` approach.
> Can we do it in a later PR if it turns out to be a requirement for someone? (I actually doubt it's useful to anyone other than a local development environment)



> > ah ok, if you want to sed the default config file this is fine for me. For the Arch package I do exactly this:
> 
> OK, updated using sed.
> 
> > Also creating directories inside /var/lib/sftpgo in the image should be useless if the user mount /var/lib/sftpgo.
> 
> Right, but it's explicit and works OOTB if you don't mount anything.
> 
it should work OOTB even if you don't create directories

> > uid/gid 1000 should be fine most of the time, env vars to allow to customize uid/gid could be useful too
> 
> That would complicate things, it would require an entrypoint, we might not even be able to use the current `USER` approach.
> Can we do it in a later PR if it turns out to be a requirement for someone? (I actually doubt it's useful to anyone other than a local development environment)

well if you want to run it as systemd service you have to use 1000 as uid/gid, however we could add an entrypoint later

sagikazarmark:
> well if you want to run it as systemd service you have to use 1000 as uid/gid, however we could add an entrypoint later

Maybe it's just me, but I don't believe in running docker containers with systemd :P

Is there anything else you need me to change?

drakkan:
I pushed the image and changed its visibility to public, I also configured CR_PAT, can you please uncomment the relevant section so will see it it works before pushing? Thank you

drakkan (approved):

sagikazarmark:
Done

drakkan:
thank you! If you want help with the docs this will be very appreciated too

drakkan:
Do you know why the description is splitted in multiple labels?

```
--label org.opencontainers.image.description=Fully featured and highly configurable SFTP server with optional FTP/S and WebDAV support. It can serve local filesystem --label S3 --label GCS
```

sagikazarmark:
> thank you! If you want help with the docs this will be very appreciated too

I'll try to come up with something, but probably not until I start using it. I will also create a helm chart, I'd definitely love to contribute it back. You can easily host a helm chart repository using github actions:

- https://github.com/helm/chart-releaser-action
- https://github.com/pomerium/pomerium-helm

> Do you know why the description is splitted in multiple labels?

no clue. Maybe it's too long or multiline?

sagikazarmark:
> Maybe it is the comma

Oh, yeah, that's absolutely possible. The docker action splits by lines and commas as well.

You could either shorten/remove commas, or replace the commas during in the build.

I can provide a PR if you choose the latter.

drakkan:
> > Maybe it is the comma
> 
> Oh, yeah, that's absolutely possible. The docker action splits by lines and commas as well.
> 
> You could either shorten/remove commas, or replace the commas during in the build.
> 
> I can provide a PR if you choose the latter.

I simply changed the description inside the container. I don't think I'll change SFTPGo description so often

drakkan:
> > I'll evaluate to add an CLI flag but it would be disabled by default since most users want to avoid manual configuration as much as possible
> 
> As long as I have control over it, I don't mind if I have to manually disable it.

Please take a look [here](https://github.com/drakkan/sftpgo/commit/c992072286be8657ab8cafc7fa2a84f80838b2a8)

sagikazarmark:
Thanks!

sagikazarmark:
Although, on second look, a boolean flag, like `enable_auto_init` (disable, depends on the default value) or something like that would have been better IMHO.

drakkan:
> Although, on second look, a boolean flag, like `enable_auto_init` (disable, depends on the default value) or something like that would have been better IMHO.

I evaluated that and it could be easily changed to work this way, but I see no clear advantage, I avoided to pass a new parameter. The configuration object is easily to extend.

drakkan:
Hi, I did some minor changes to the Dockerfile and added support for building for arm64 too. I wrote some basic [docs](https://github.com/drakkan/sftpgo/tree/master/docker) feel free to improve them, thank you!

sagikazarmark:
Cool! Docs look great. I had one comment about the order of instructions in the Dockerfile though:

https://github.com/drakkan/sftpgo/commit/87f78b07b3bfc88bb821525f478f2dc8742c6a26#r43002271

drakkan:
I'll change the order thank you. 

Do you have idea why the labels are empty here?

https://github.com/users/drakkan/packages/container/sftpgo/111582

for example: `"org.opencontainers.image.url": "",`

maybe the github context is different while running on schedule? For the edge version the description is not yet available, maybe it will be populated after some time, let's if the label are empty there too.

sagikazarmark:
Yeah, it looks like the `${{ github.event.repository.html_url }}` context is missing when an image is built.

Reported the issue here: https://github.com/docker/build-push-action/issues/165

drakkan:
Hi, 

I plan to re-add build args like the previous [TAG and FEATURES](https://github.com/drakkan/sftpgo/blob/master/docker/sftpgo/alpine/Dockerfile#L6) to allow to customize the image if someone want to build it itself, and maybe add a debian based Dockerfile too.

Probably the current COMMIT_SHA arg was intented to work as the previous TAG build arg.

 Do you have other suggestions?

sagikazarmark:
I can't really see what `TAG` was used for, but it's not entirely the same as `COMMIT_SHA` (looking at the file you linked).
There shouldn't be any git checkout commands inside the build.

As long as the default for `FEATURES` is to build everything, 👍

drakkan:
> I can't really see what `TAG` was used for, but it's not entirely the same as `COMMIT_SHA` (looking at the file you linked).

TAG was used to specify a git commit and/or a tag to build, see #105

Oh, yes `COMMIT_SHA` is a different things it is used in docker.yml

> There shouldn't be any git checkout commands inside the build.
> 

why not? I'm not a docker expert :)

> As long as the default for `FEATURES` is to build everything,

Yes, all by default

sagikazarmark:
> why not? I'm not a docker expert :)

If you want to build a specific revision, you should check that revision out and run docker build there.

drakkan:
> > why not? I'm not a docker expert :)
> 
> If you want to build a specific revision, you should check that revision out locally and run docker build there.

Yes this could be made manually, before the code was downloaded inside the Dockerfile, 

so basically only the build arg to allow to customize the features should be restored, thank you!

drakkan:
Hello @sagikazarmark 

thank you again for your contributions. We have recently updated our contribution policy and we now require signing a CLA. Agreeing to the CLA explicitly states that you are entitled to provide a contribution and that you cannot withdraw permission to use your contribution at a later date. This removes any ambiguities or uncertainties.

https://cla-assistant.io/drakkan/sftpgo?pullRequest=180

Thank you!

sagikazarmark:
@drakkan signed

drakkan:
> @drakkan signed

Thank you!

PR #187, opened 2020-10-08
Fix empty env context in nightly builds

sagikazarmark:
Based on https://github.com/docker/build-push-action/issues/165#issuecomment-705558902


drakkan:
thank you, let's see if this fix the descrition not available even for the edge image

PR #200, opened 2020-10-19
Add support for custom action hooks

### LINKED ISSUE #186: Custom SSH proxy in portable mode with custom hooks
I'd like embed sftpgo into a custom service and start it in portable mode, so it can basically act as an sftp proxy.

I'd also like to implement a custom hook, that sends a message to a queue (I'd rather not implement it as a process or an HTTP endpoint because of retries, etc)

Is this a supported use case at the moment? Can it even be supported? Tried to find the right extension points, but couldn't so far.

sagikazarmark:
This PR adds support for the feature explained in #186

The PR only adds an interface and refactors the minimal code necessary. Further refactoring might make sense in a subsequent PR.

Fixes #186

### DIFF
--- a/common/actions.go
+++ b/common/actions.go
@@ -34,8 +34,29 @@ type ProtocolActions struct {
 	Hook string `json:"hook" mapstructure:"hook"`
 }
 
-// actionNotification defines a notification for a Protocol Action
-type actionNotification struct {
+var actionHandler ActionHandler = defaultActionHandler{}
+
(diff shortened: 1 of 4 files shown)


drakkan:
Hi,

the patch looks good, thank you! I would just add a little test case to show how to use ʻInitializeActionHandler` so codecov will be happy too

sagikazarmark:
Added an extra test for that function.

drakkan:
Thank you!

PR #201, opened 2020-10-20
Migrate to WithCredentialsJSON for GCS

Deiz:
The current WithCredentialsFile approach leaves state in Docker
containers, on disk, etc.

The Google Cloud SDKs support a WithCredentialsJSON option provider that
reads from a byte slice.

I've also changed the credentials format to be []byte and removed the
manual base64 decoding and encoding, as encoding/json wrangles that
automatically.

Support for existing credentials files has been preserved, but they are
no longer written to disk when adding or updating users, and instead go
into the dataprovider as part of the user.


drakkan:
Hi,

thank you for your contribution, very appreciated.

The credentials files are, by default, in `/var/lib/sftpgo`. This folder contains host keys, database file (if you use SQLite or bolt), eventually a configuration file etc. It is the config dir and it is declared as volume. 

You can also change the location for the credentials dir using an env var.

With this change if you list users, using for example the REST API, the GCS credentials are fetched from the data provider and then discarded, since they contain sensitive data that should not be exposed.

The current approach loads the credential files only when a dump is requested, so there is less load on the provider.

This is the classical "should I store files inside the database or in FileSystem" debate, I need some time to better evaluate this PR. Thank you

Deiz:
I'm thinking about the Google service account the same way as the AWS access key ID & secret access key, with the latter two already being stored in the dataprovider, so my view is that this PR just makes the S3 and GCS filesystems more consistent with each other.

That said, happy to compromise here; maybe we can support both with either a dataprovider config flag (`prefer_database_credentials`?) or a GCS filesystem option that could be passed when creating/updating users, which might be supplied via `--gcs-store-credentials-in-database` on the CLI?

If it helps, my use case is follows:

By baking a host key into the image, using Postgres as the dataprovider, and GCS as the filesystem, I'm avoiding having SFTPGo create any disk state at runtime. Being able to read the Google Cloud service account data from the dataprovider is the only thing that still creates files on disk with that configuration.

dataprovider/bolt.go
```diff
@@ -352,6 +352,7 @@ func (p BoltProvider) updateUser(user User) error {
 		user.UsedQuotaSize = oldUser.UsedQuotaSize
 		user.UsedQuotaFiles = oldUser.UsedQuotaFiles
 		user.LastLogin = oldUser.LastLogin
+		user.ID = oldUser.ID
```
drakkan, line 355: Is this required? If so we should add it to the memory provider too. This case should be checked [here](https://github.com/drakkan/sftpgo/blob/master/httpd/api_user.go#L141) already. Is there some failing test case?

drakkan:
> I'm thinking about the Google service account the same way as the AWS access key ID & secret access key, with the latter two already being stored in the dataprovider, so my view is that this PR just makes the S3 and GCS filesystems more consistent with each other.
> 
> That said, happy to compromise here; maybe we can support both with either a dataprovider config flag (`prefer_database_credentials`?) or a GCS filesystem option that could be passed when creating/updating users, which might be supplied via `--gcs-store-credentials-in-database` on the CLI?
> 

This is probably the best solution, thanks.

Maybe we can just accept an empty "credentials_path" and reuse the existing setting, documenting this behaviour. What do you think about?

> If it helps, my use case is follows:
> 
> By baking a host key into the image, using Postgres as the dataprovider, and GCS as the filesystem, I'm avoiding having SFTPGo create any disk state at runtime. Being able to read the Google Cloud service account data from the dataprovider is the only thing that still creates files on disk with that configuration.

this is a valid use case, thank you for sharing it

dataprovider/bolt.go
Deiz: This was causing most of the modified InvalidFs tests to fail because they do this:

1. Add user
2. Update user (based on add user response)
3. Delete user (based on add user response)

The flow for the Bolt data provider was:
1. Add user is fine, returns a non-zero user ID
2. Update user is called with a valid user, but due to this issue, stores a user with ID 0 at the correct user ID's key.
3. Delete user 500s because the HTTP handler uses a user ID path parameter, looks the user up, and then passes it to Bolt's `deleteUser` which does this:
```
        userIDAsBytes := itob(user.ID)
        userName := idxBucket.Get(userIDAsBytes)
```

At a glance, the memory provider doesn't suffer from this because it re-uses the passed-in user (preserving the user ID by default) where Bolt makes a new user struct instance based on the previous one.

drakkan: This likely happen because inside the tests you update the user this way:

```
err = dataprovider.UpdateUser(u)
```

without using the (public) REST API. It should not happen (or fail before) if you use `httpd.UpdateUser(....`. I'll do some tests myself later today, thank you

ftpd/ftpd_test.go
```diff
@@ -860,17 +859,29 @@ func TestLoginInvalidFs(t *testing.T) {
...
 	}
+
+	// write credentials to a file to trigger legacy behaviour
+	err = ioutil.WriteFile(credentialsFile, u.FsConfig.GCSConfig.Credentials, 0600)
+	assert.NoError(t, err)
+
+	// remove dataprovider credentials to force credentials file fallback
+	u.FsConfig.GCSConfig.Credentials = nil
+	err = dataprovider.UpdateUser(u)
+	assert.NoError(t, err)
```
drakkan, line 879: Please use: 

```
user.FsConfig.GCSConfig.Credentials = nil
user, _, err = httpd.UpdateUser(user, http.StatusOK, "")
```

Deiz: Done

drakkan:
Merged, thank you!

drakkan:
Hi @Deiz 

[this](https://github.com/drakkan/sftpgo/commit/6ad4cc317cc0c3afa685f7645c84e21f1cd00a9e) commit should improve the performance of the cloud backends. I would be really grateful if you could try it out and confirm that no bugs have been introduced, thanks!

drakkan:
Hello,

https://cla-assistant.io/drakkan/sftpgo?pullRequest=201

The contributed code has been rewritten but we would still greatly appreciate it if you could sign the CLA.
Thank you!

drakkan:
Hello @Deiz 

https://cla-assistant.io/drakkan/sftpgo?pullRequest=201

The contributed code has been rewritten but we would still greatly appreciate it if you could sign the CLA.
Thank you!

PR #213, opened 2020-10-29
Add health check endpoint

### LINKED ISSUE #212: Add a healthz endpoint and exclude it from logs and metrics
Kubernetes (and many other orchestrators) require a health endpoint that it can use to check if the application is still running and responding to requests.

Kubernetes usually hits these endpoints every few seconds, so it generates a large traffic that can clutter logs and metrics. 

The closes thing I found is a version endpoint that does nothing, but returns a successful response: `/api/v1/version`

I'd be happy to provide a PR, but wanted to raise an issue first.

sagikazarmark:
Fixes #212


drakkan:
Thank you

PR #222, opened 2020-11-11
update user.html

ryjogo:
add relative


drakkan:
Hi, thank you for the pull request but maybe we should find a better term. This is a SFTP/FTP/HTTP path in UNIX format. Windows users could get confused from the `relative` term and write something like `\dir\subdir`. I'm not very good at finding the right terms, can you please suggest some alternatives? Maybe "One SFTP/FTP directory path per line ...", "One virtual directory path per line ..." or "One directory path, as seen by the user, per line .."?

ryjogo:
Ah yes OK. Perhaps:

"One virtual directory path per line as dir::perms, for example /somedir::list,download* 

 is fine.

drakkan:
yes please change this way, thank you

drakkan:
merged, thank you

Issue #229, opened 2020-11-23
Memory data provider sometimes requires extra configuration


estelsmith:
When configuring the `memory` data provider, SFTPGo will fail to start by default on read-only filesystems, such as in an unprivileged container.

Example:

```
data_provider:
    driver: "memory"
```

Logs:

```
{"level":"debug","time":"2020-11-23T22:18:24.658","sender":"dataprovider_memory","message":"loading users from file: \"/sftpgo.db\""}
{"level":"warn","time":"2020-11-23T22:18:24.659","sender":"dataprovider_memory","message":"error loading users: stat /sftpgo.db: no such file or directory"}
{"level":"error","time":"2020-11-23T22:18:24.659","sender":"service","message":"error initializing data provider: stat /sftpgo.db: no such file or directory"}
```

---

I've found that this was being caused by the `name` configuration option having a default value, so setting `name=""` allows it to work.

It's not a huge issue, but it does cause unexpected errors that need tracked down. I'd imagine that setting the `memory` data provider would require no further configuration, as there shouldn't be a `name`d file to load data from unless I explicitly specify one.

Thoughts?

estelsmith:
The same can be said about the `credentials_path` data provider option. It will cause SFTPGo to fail on read-only filesystems by default.

I'd think that not providing a `credentials_path` would mean that SFTPGo wouldn't attempt to touch the filesystem.

drakkan:
Hi,

for credentials please try to set `prefer_database_credentials` to `true` this way SFTPGo should ignore the `credentials_path`.

The `name` setting allows you to configure the file that the memory provider will use to load users and folders, so we can't automatically assume it as empty, you have to configure it explicitly sorry

estelsmith:
@drakkan Would it be preferable to default the field to empty if `memory` is chosen? That way if you specify the memory provider, it won't automatically fail in read-only environments, but still allow users to specify a file to load data from?

I feel this would strike the best balance between user-friendliness and having a default `name` set for other providers that actually require it to function.

estelsmith:
@drakkan The same maybe could be said about the `credentials_path`? Perhaps `prefer_database_credentials` could default to `true`, only deferring to the filesystem if someone explicitly configures it to do so?

I'm not sure what's the best way for it, I was just thinking of how to make it the simplest to use without unexpected errors jumping out at the user.

drakkan:
Hi,

should I provide a different default config file for each provider?

[...]

PR #231, opened 2020-11-24
Add ability to conditionally set default data_provider.name

estelsmith:
To fix an issue reported in #229.

We now have `config.setConditionalDefaults()` called after the configuration file is parsed. This allows us to change default configurations based on any criteria desired.

Currently it is being used to set `data_provider.name` to `""` by default if using the `memory` driver.

This is only a default and can be overridden by any available configuration methods.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -154,6 +154,7 @@ jobs:
             org.opencontainers.image.licenses=${{ fromJson(steps.repo.outputs.result).license.spdx_id }}
 
       - name: Build and push full
+        if: ${{ github.event_name != 'pull_request' }}
         uses: docker/build-push-action@v2
         with:
           builder: ${{ steps.builder-full.outputs.name }}
--- a/cmd/startsubsys.go
+++ b/cmd/startsubsys.go
@@ -72,7 +72,7 @@ Command-line flags should be specified in the Subsystem declaration.
 				logger.Debug(logSender, connectionID, "data provider %#v not supported in subsystem mode, using %#v provider",
 					dataProviderConf.Driver, dataprovider.MemoryDataProviderName)
 				dataProviderConf.Driver = dataprovider.MemoryDataProviderName
-				dataProviderConf.Name = ""
+				dataProviderConf.Name = new(string)
 				dataProviderConf.PreferDatabaseCredentials = true
 			}
 			config.SetProviderConf(dataProviderConf)
--- a/common/common_test.go
+++ b/common/common_test.go
@@ -166,9 +166,14 @@ func initializeDataprovider(trackQuota int) (string, error) {
 	if err := viper.ReadInConfig(); err != nil {
 		return "", err
 	}
+	// @TODO Why doesn't this use the config loader? It bypasses setting conditional defaults
 	var cfg providerConf
 	if err := viper.Unmarshal(&cfg); err != nil {
 		return "", err
+	}
+	if cfg.Config.Name == nil {
+		name := "sftpgo.db"
+		cfg.Config.Name = &name
 	}
 	if trackQuota >= 0 && trackQuota <= 2 {
 		cfg.Config.TrackQuota = trackQuota
--- a/config/config.go
+++ b/config/config.go
@@ -45,7 +45,7 @@ type globalConfig struct {
 	HTTPConfig   httpclient.Config     `json:"http" mapstructure:"http"`
 }
 
-func init() {
+func Init() {
 	// create a default configuration to use if no config file is provided
 	globalConf = globalConfig{
 		Common: common.Configuration{
@@ -115,7 +115,7 @@ func init() {
 		},
 		ProviderConf: dataprovider.Config{
 			Driver:           "sqlite",
-			Name:             "sftpgo.db",
+			Name:             nil,
 			Host:             "",
 			Port:             5432,
 			Username:         "",
@@ -173,6 +173,9 @@ func init() {
 	setViperDefaults()
 	viper.AutomaticEnv()
 	viper.AllowEmptyEnv(true)
+}
+func init() {
+	Init()
 }
 
 // GetCommonConfig returns the common protocols configuration
@@ -283,6 +286,7 @@ func LoadConfig(configDir, configName string) error {
 		logger.WarnToConsole("error parsing configuration file: %v. Default configuration will be used.", err)
 		return err
 	}
+	setConditionalDefaults()
 	checkCommonParamsCompatibility()
 	if strings.TrimSpace(globalConf.SFTPD.Banner) == "" {
 		globalConf.SFTPD.Banner = defaultSFTPDBanner
@@ -325,6 +329,20 @@ func LoadConfig(configDir, configName string) error {
 	checkHostKeyCompatibility()
 	logger.Debug(logSender, "", "config file used: '%#v', config loaded: %+v", viper.ConfigFileUsed(), getRedactedGlobalConf())
 	return err
+}
+
+func setConditionalDefaults() {
+	var conf = &globalConf.ProviderConf
+
+	if conf.Name == nil {
+		var defaultName = "sftpgo.db"
+
+		if conf.Driver == "memory" {
+			defaultName = ""
+		}
+
+		conf.Name = &defaultName
+	}
 }
 
 func checkHostKeyCompatibility() {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -329,3 +329,96 @@ func TestConfigFromEnv(t *testing.T) {
 	assert.Len(t, dataProviderConf.Actions.ExecuteOn, 1)
 	assert.Contains(t, dataProviderConf.Actions.ExecuteOn, "add")
 }
+
+// Ensure that "data_provider.name"'s default value is changed based on "data_provider.driver"
+func TestConditionalDefaultProviderName(t *testing.T) {
+	// Temporarily unset environment variables, because they supersede other configurations.
+	envs := []string{
+		"SFTPGO_DATA_PROVIDER__NAME",
+		"SFTPGO_DATA_PROVIDER__DRIVER",
+	}
+
+	for i := range envs {
+		(func(name string) {
+			currentVal, exists := os.LookupEnv(name)
+
+			if exists {
+				os.Unsetenv(name)
+
+				t.Cleanup(func() {
+					os.Setenv(name, currentVal)
+				})
+			}
+		})(envs[i])
+	}
+
+	configDir := ".."
+	confName := tempConfigName + ".json"
+	configFilePath := filepath.Join(configDir, confName)
+
+	type TestProvider struct {
+		Driver string  `json:"driver"`
+		Name   *string `json:"name"`
+	}
+	type TestJSON struct {
+		DataProvider TestProvider `json:"data_provider"`
+	}
+	loadConfig := func(conf TestJSON) {
+		serializedConfig, err := json.Marshal(conf)
+		assert.NoError(t, err)
+
+		err = ioutil.WriteFile(configFilePath, serializedConfig, os.ModePerm)
+		assert.NoError(t, err)
+
+		config.Init()
+
+		err = config.LoadConfig(configDir, tempConfigName)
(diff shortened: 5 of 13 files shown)


estelsmith:
Boo... I need to fix what appear to be issues loading the database name for postgres and mysql, and include tests to cover the additional code.

drakkan:
Hi, the docker build for pull request should be fixed in the latest commit.

This time it's me who doesn't understand the documentation :)

I'll test this patch later but I think that the only way to have a nil name is to remove the name key from the configuration file. Am I wrong?

estelsmith:
> I'll test this patch later but I think that the only way to have a nil name is to remove the name key from the configuration file. Am I wrong?

You can either set it as `null`, or omit it entirely. If you set the value in the configuration file, then that's what it will be set to.

I'm getting ready to push a change for tests, so keep an eye out for that.

drakkan:
> > I'll test this patch later but I think that the only way to have a nil name is to remove the name key from the configuration file. Am I wrong?
> 
> You can either set it as `null`, or omit it entirely. If you set the value in the configuration file, then that's what it will be set to.
> 

the default configuration should contain all supported (and not deprecated) values, so please set the name to `null` in sftpgo.json,

I will review this PR after dinner, thank you!

> I'm getting ready to push a change for tests, so keep an eye out for that.

drakkan:
If I correctlt understand what you mean, this way we should have a default config like this:

```
 "driver": "sqlite",
 "name": null,
```

so by default a user should read the docs to understand that for sqlite/bolt the default database is `sftpgo.db`.

If we leave the configuration as is, for your use case you must set the name to an empty string to `null` or remove it. 

I'm still not entirely convinced that this will improve things. It seems that this way we create confusion by default, while before we create confusion if someone change the setting

estelsmith:
@drakkan 

> If I correctlt understand what you mean, this way we should have a default config like this:

I updated the default configuration at `sftpgo.json`:

```
"driver": "sqlite",
"name": null
```

It doesn't matter if `name` is set to `null` or simply discarded, it will still read the default value.

With this PR, `name` will be whatever the user sets it to, defaults to `""` for memory provider, or the original default `"sftpgo.db"`.

I do think that a documentation update should likely accompany this PR, so they know what to expect this value to be. Current documentation doesn't even tell you what the default value is, only that it's "optional" if you use the memory provider (which is patently untrue if using a readonly FS).

drakkan:
> @drakkan
> 
> > If I correctlt understand what you mean, this way we should have a default config like this:
> 
> I updated the default configuration at `sftpgo.json`:
> 
> ```
> "driver": "sqlite",
> "name": null
> ```
> 
> It doesn't matter if `name` is set to `null` or simply discarded, it will still read the default value.
> 
> With this PR, `name` will be whatever the user sets it to, defaults to `""` for memory provider, or the original default `"sftpgo.db"`.
> 
> I do think that a documentation update should likely accompany this PR, so they know what to expect this value to be. Current documentation doesn't even tell you what the default value is, only that it's "optional" if you use the memory provider (which is patently untrue if using a readonly FS).

is optional means you can omit it, or at least it was the intended meaning :) 

I'll think about this pull request, please rebase against https://github.com/drakkan/sftpgo/commit/ff73e5f53c8474e21fc7eebe51d2bbd8687e7b01 so Docker build should succeed, test cases seems to fail when using bolt provider (the database name is sftpgo_bolt.db for bolt test cases) thank you!

estelsmith:
> is optional means you can omit it, or at least it was the intended meaning :)

Good because that's my intent with this PR. You should be able to omit the `name` field without causing any side-effects, even in environments with a readonly FS. (yes, that's my primary case)

> I'll think about this pull request, please rebase against ff73e5f so Docker build should succeed, test cases seems to fail when using bolt provider (the database name is sftpgo_bolt.db for bolt test cases) thank you!

Sure! I'm still tracking down a couple of issues that CI is catching. I'll make a comment once it's complete.

estelsmith:
@drakkan This should be complete now, provided the tests pass.

If you have an issue with any of it, let me know and I'll look into it.

drakkan:
@estelsmith Thanks the patch looks technically correct. I'll take a deeper look tomorrow.

To summarize, before this patch we had a configuration like this:

```
"driver": "sqlite",
"name": "sftpgo.db",
```

now we have:

```
"driver": "sqlite",
"name": null,
```

and name is automatically set to "" if driver is `memory`, otherwise to `sftpgo.db`.
This prevents sftpgo from not starting if someone has set the driver to `memory` without also setting the provider name to an empty string. 

I'm a bit uncertain here.

@sagikazarmark I know, I'm testing your patience by asking your opinion on multiple issues and I fully understand if you don't want or don't have time to answer: as a viper maintainer and therefore an authoritative voice for the configuration systems, what do you think about this proposed change? Does anyone else want to express their opinion?

drakkan:
a, minor, issue is that the provider name not printed in logs anymore:

```
ProviderConf:{Driver:sqlite Name:0xc00016bc40
```

we should implement the Stringer interface for the config struct if we want to log it again

sagikazarmark:
@drakkan no problem. 🙂 As I always tell my grandmother: feel free to ping (call) me, I'll simply not answer if I'm busy. 😄 

TBH, I'm not sure what this change does. Does the memory provider break if there is a name defined? I'd probably fix that instead, sounds much easier to me.

Personally, I think SFTPGo's configuration is already complex as it is (mostly because of the global nature of it) and this change adds to that complexity, making the code less obvious, thus more error prone to change it.

I could think of a number of improvements on the configuration front in SFTPGo that could potentially solve the original issue as well:

- don't make the configuration global: inject it
- use URL-like DSNs for data providers (eg. `sqlite://sqlite.db`, `memory://`) (that would solve this problem as well)
- if you don't want DSNs, unmarshal config into different structs based on the driver name (mapstructure can help with that) or simply don't process certain arguments for drivers

**tl;dr** it feels like this is a complicated patch for a problem that could be fixed with a various other improvements. That if I understood correctly what this patch addresses.

drakkan:
> @drakkan no problem.  As I always tell my grandmother: feel free to ping (call) me, I'll simply not answer if I'm busy. 
> 
> TBH, I'm not sure what this change does. Does the memory provider break if there is a name defined? I'd probably fix that instead, sounds much easier to me.
> 

if you define a name for the memory provider it expects a user dump. If the file is not found or if it is not a valid dump SFTPGo will not start, maybe we could  be more forgiving, print a warning and start anyway in this case

> Personally, I think SFTPGo's configuration is already complex as it is (mostly because of the global nature of it) and this change adds to that complexity, making the code less obvious, thus more error prone to change it.
> 
> I could think of a number of improvements on the configuration front in SFTPGo that could potentially solve the original issue as well:
> 
> * don't make the configuration global: inject it

this is a good point
> * use URL-like DSNs for data providers (eg. `sqlite://sqlite.db`, `memory://`) (that would solve this problem as well)

This is already possibile using `data_provider.connection_string`, this string must be the DSN expected by the upstream provider. It is available for any provider != memory

> * if you don't want DSNs, unmarshal config into different structs based on the driver name (mapstructure can help with that) or simply don't process certain arguments for drivers
> 

the `name` parameter is useful for `memory` provider, however I have to study this solution better

> **tl;dr** it feels like this is a complicated patch for a problem that could be fixed with a various other improvements. That if I understood correctly what this patch addresses.

drakkan:
@estelsmith I would like to fix #229 with a patch like this one:

```
diff --git a/dataprovider/memory.go b/dataprovider/memory.go
index 923f36b..5c1a26f 100644
--- a/dataprovider/memory.go
+++ b/dataprovider/memory.go
@@ -61,7 +61,10 @@ func initializeMemoryProvider(basePath string) error {
                        configFile:    configFile,
                },
        }
-       return provider.reloadConfig()
+       if err := provider.reloadConfig(); err != nil {
+               logger.ErrorToConsole("unable to load initial data: %v", err)
+       }
+       return nil
 }
 
 func (p MemoryProvider) checkAvailability() error {
```

since you did a lot of work here I would like to push this commit using your name/email as author, do you agree?

I'll probably push a similar commit to do the same for the `--loaddata-from` serve flag

estelsmith:
@drakkan @sagikazarmark 

> use URL-like DSNs for data providers (eg. sqlite://sqlite.db, memory://) (that would solve this problem as well)

👍 I would love to see something like this implemented in lieu of the list of different options presented in the provider configuration.

> I would like to fix #229 with a patch like this one:

Actually, that's a lot simpler for sure. It does fix the issue because SFTPGo won't exit if the file doesn't exist.

> since you did a lot of work here I would like to push this commit using your name/email as author, do you agree?

Hard pass. I appreciate the sentiment, but I didn't author that patch. You ought to take full credit for it!!

---

Since there's been sentiment expressed against the configuration system, should we reject this PR and open an issue to address its shortcomings?

drakkan:
> @drakkan @sagikazarmark
> 
> > use URL-like DSNs for data providers (eg. sqlite://sqlite.db, memory://) (that would solve this problem as well)
> 
>  I would love to see something like this implemented in lieu of the list of different options presented in the provider configuration.
> 

This should be already possibile for every provider != memory using the `connection_string` configuration. I could add support for the memory provider too, it should be quite easy. For me this is more difficult than filling the splitted fields (you are forced to find/read some docs) but this is just my opinion

> > I would like to fix #229 with a patch like this one:
> 
> Actually, that's a lot simpler for sure. It does fix the issue because SFTPGo won't exit if the file doesn't exist.
> 
> > since you did a lot of work here I would like to push this commit using your name/email as author, do you agree?
> 
> Hard pass. I appreciate the sentiment, but I didn't author that patch. You ought to take full credit for it!!
> 

are you sure? I have these pending commits that should fix all the issues you raised

```
commit 2054dfd83d89bc044cf9c533886f0e016e3c2ed2 (HEAD -> master)
Author: Nicola Murino <nicola.murino@gmail.com>
Date:   Wed Nov 25 14:18:12 2020 +0100

    create the credential directory when needed
    
    The credentials dir is currently required only for GCS users if
    prefer database credential setting is false, so defer its creation
    and don't fail to start the services if this directory is missing

commit 6699f5c2cc88c86f0b735c7556ff549652ca74b8
Author: Nicola Murino <nicola.murino@gmail.com>
Date:   Wed Nov 25 09:18:36 2020 +0100

    initial data loading: an error is no longer fatal
    
    therefore it does not prevent the services from starting

commit 70bde8b2bc8f7cb77e4f9bdc924619957eecb8b9
Author: Estel Smith <estel.smith@gmail.com>
Date:   Wed Nov 25 09:15:23 2020 +0100

    memory provider: print a log if loading the initial dump fails
    
    therefore this error is no longer fatal and does not prevent the services
    from starting
    
    Fixes #229
```

I would really like to give you the attribution for the first patch

> Since there's been sentiment expressed against the configuration system, should we reject this PR and open an issue to address its shortcomings?

For the config we have a mixed approach: 

- a global config 
- for each service only the relevant part of the global config is injected

If we remove the global config we'll probably break the portable mode and we have to refactor some things in the dataprovider. The dataprovider was the first code I wrote and I was just starting to learn Go at that time, so I think it could be improved a lot.

I have other priorities for now, but if you want to spend some time to identify what to improve please open a new issue. 

I'm working on these new features:

- kms support. I'm evaluating different kms libraries and external bridge like KES. The KMS will be used to decrypt cloud backend credentials for now
- sftp fs backend
- cryptofs backend (for local fs)
- a swift storage backend (maybe, I'm not sure here)

both sftp fs and crypto fs will use kms if configured. I could play with the configuration system after finishing these things.

estelsmith:
> I would really like to give you the attribution for the first patch

Okay, I won't fight it. 👍 

> If we remove the global config we'll probably break the portable mode and we have to refactor some things in the dataprovider. The dataprovider was the first code I wrote and I was just starting to learn Go at that time, so I think it could be improved a lot.

Yeah, I notice the global config is sprinkled around a couple of spots, but it's especially gnarly dealing with it in tests. It's why my test case became so large dealing with getting/keeping a clean state.

> I have other priorities for now, but if you want to spend some time to identify what to improve please open a new issue.

Sure. I'll take a look and open an issue when I feel I have something more substantial.

PR #237, opened 2020-12-02
Accept a path in config file flag

### LINKED ISSUE #236: Simplify configuration file/dir
Currently configuring SFTPGo is a bit complicated in the various environments it might run in (thanks to the available Docker image).

Namely, the fact that the config dir is tied to a bunch of other things (host keys, sqlite, etc) and that you cannot pass a full path to a config file is a bit uncomfortable.

To give you a concrete example: when running on Kubernetes, I can mount a config map to a certain path (eg. `/etc/sftpgo`), containing an `sftpgo.yaml` file. The problem is, by default, SFTPGo will try to write default host keys to the config dir, but that mount is read only. Right now I solve this by manually setting the path to host keys.

I see two potential improvements:

- allow passing a full path to SFTPGo as a flag (analogous to [`viper.SetConfigFile`](https://pkg.go.dev/github.com/spf13/viper#SetConfigFile))
- introduce a separate directory for host keys and other stuff (although there are already enough directories)

The first approach would allow mounting a config map under something like `/etc/sftpgo/config` (and since `--config-file` is already taken it would allow passing `--config-path /etc/sftpgo/config/sftpgo.yaml`) and would not interfere with the current config dir concept (`/etc/sftpgo` would still be writable).

sagikazarmark:
This PR is a major breaking change to how sftpgo accepts configuration files.

So far the `--config-file` flag (or `SFTPGO_CONFIG_FILE` env var) accepted a so called "config name", a file name without extension. Internally, Viper (the config library used under the hood) uses this config name to find a config file in various paths with various extensions.

This is not really useful to users though. They usually want to pass an known config file to sftpgo instead of a kind of search parameter.

This PR changes the behavior of the above mentioned flag (and env var): they accept an absolute path to a config file or a relative path (to the working directory of the process).

Fixes #236

### DIFF
--- a/cmd/install_windows.go
+++ b/cmd/install_windows.go
@@ -64,7 +64,7 @@ func getCustomServeFlags() []string {
 		result = append(result, "--"+configDirFlag)
 		result = append(result, configDir)
 	}
-	if configFile != defaultConfigName {
+	if configFile != "" {
 		result = append(result, "--"+configFileFlag)
 		result = append(result, configFile)
 	}
--- a/cmd/portable.go
+++ b/cmd/portable.go
@@ -124,7 +124,6 @@ Please take a look at the usage below to customize the serving parameters`,
 			}
 			service := service.Service{
 				ConfigDir:     filepath.Clean(defaultConfigDir),
-				ConfigFile:    defaultConfigName,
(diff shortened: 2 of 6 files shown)


drakkan:
thank you

PR #247, opened 2020-12-11
Update sftpgo_api_cli

dharmendrakariya:


drakkan:
Are you reverting api cli in git master to 1.2.2? This way you will break git master. This version of api cli will not work with edge docker image variant. 

If you are using sftpgo v1.2.2 you should use the same version for the api cli too.

dharmendrakariya:
I am using helm chart and in that docker image (not the git image) didn't know the version.

sorry I didn't pay attention to your reply. anyway I think docker image supposed to be the latest one in this case.

drakkan:
Hi, the helm chart is the latest **stable** version

https://artifacthub.io/packages/helm/sagikazarmark/sftpgo

so it is v1.2.2. You cannot use the development version of the rest api cli against the latest stable version, for example in git master we have several features that are not inside v1.2.2 (kms support, data at-rest encryption, status api etc..).

The development version of the API CLI works with the development version of sftpgo.

If you use the docker image directly it depends on the tag you use, more info here:

https://github.com/drakkan/sftpgo/tree/master/docker

thank you

P.S. the sftpgo version is printed inside the logs at startup and as footer inside the built-in web admin

Issue #253, opened 2020-12-17
Bind SFTP Server to Multiple Ports?


valdearg:
Can the SFTP server bind to multiple ports? I like to bind to default port 22 and a higher port just in case it's blocked elsewhere.

sagikazarmark:
No, it cannot and I don't think it would be trivial to implement (without nasty hacks like a goroutine merging connections).

IMHO it's quite standard that the operator has to choose an appropriate port.

drakkan:
I think we could try to change the configuration from this:

```
"sftpd": {
    "bind_port": 2022,
    "bind_address": "",
    ...
```

to something like this:

```
"sftpd": {
    "bindings": [
      {
        "port": 2022,
        "address": ""
      }
    ],
    ....
```

so you can add other bindings, for example:

```
 "sftpd": {
    "bindings": [
      {
        "port": 2022,
        "address": ""
      },
      {
        "port": 22,
        "address": "192.168.1.1"
      },
      {
        "port": 2222,
        "address": "192.168.1.10"
      }
    ],
    ...
```

for each binding we'll start a new listener, instead of a [single listener](https://github.com/drakkan/sftpgo/blob/master/sftpd/server.go#L184)

sagikazarmark:
I honestly haven't seen any other software allowing to bind a service to multiple ports (except maybe webservers, but it does make sense in that case). I'm not sure it's worth the effort.

A potential alternative is binding sftpgo to a high enough port (eg. 10022) and running a proxy server next to it that binds additional ports. (This is actually what happens when you use a loadbalancer)

valdearg:
Quite a lot of services allow it in my experience, including the regular OpenSSH Server.

sagikazarmark:
Hm, interesting. I wasn't aware of it.

drakkan:
OpenSSH allows to specify multiple `ListenAddress` this should be quite similar to the solution I proposed above, do you agree?

Multiple bindings will be allowed for SFTP and FTP services, I think a single binding is enough for REST API and telemetry services.

sagikazarmark:
I'm still not sure if it's worth the effort, I'd still go with the proxy solution, but if it's an accepted solution I guess it's okay.

drakkan:
I'm trying to implement this:

```
"sftpd": {
    "bindings": [
      {
        "address": "",
        "port": 2022,
        "apply_proxy_config": true
      }
    ],
...
```

[...]

Issue #197, opened 2020-10-18
REST API/Web admin users and permissions


drakkan:
Hi,

currently REST API and Web Admin can be protected using HTTP basic authentication via an file created using the Apache's `htpasswd` tool.

This works well for basic needs but I think it can be improved.

We can move the admins inside the data provider and add support for read and write permissions too. After the initial installation a default admin with a predefined password will be created and this admin can create other administrators and set their permissions via REST/API or Web Admin. For example you can create a new admin that can only create users but not terminate active connections and so on.

Maybe external authentication for this administrators could be added too so for example existing LDAP users or such can be used.

What do you think about? Is this feature useful for your use case? Thank you!

HiFiPhile:
Hi,

Currently in my configuration SFTPGO's Web Admin and my other services are hidden behind HAProxy, so I can centralize authentication and use Fail2Ban.

I feel like multi-administrators is a nice to have but niche for most cases. 

Maybe we can start by adding `X-Real-IP` or `X-Forwarded-For` to have real ip in logs and add a Fail2Ban config.

drakkan:
> Hi,
> 
> Currently in my configuration SFTPGO's Web Admin and my other services are hidden behind HAProxy, so I can centralize authentication and use Fail2Ban.
> 
> I feel like multi-administrators is a nice to have but niche for most cases.
> 
> Maybe we can start by adding `X-Real-IP` or `X-Forwarded-For` to have real ip in logs and add a Fail2Ban config.

Thank you for your feedback.

The real ip should be already [supported](https://github.com/drakkan/sftpgo/blob/master/httpd/router.go#L26) can you please confirm? Regarding Fail2Ban I don't know it very well a pull request would be great

HiFiPhile:
> The logs should already contain the real ip.

Indeed, that's the case :)

HiFiPhile:
Hi,

Do you think it's a good idea to add a "trusted reverse proxy" in config ? For anyone who don't use reverse proxy, source ip can be tricked by adding X-Real-IP or X-Forwarded-For.

drakkan:
Hi,

we could add something similar to HA proxy config for reverse proxy too, for example:

- ProxyProtocol
  - `0`, disabled
  - `1`, enabled. Header will be used and requests without header will be accepted
  - `2` required. 
[...]

PR #254, opened 2020-12-17
feat: add new telemetry server

### LINKED ISSUE #244: /metrics without authentication
I am using htpasswd for the mgmt UI, and i wish to scrape the metrics endpoint, i do know its possible to provide basic auth in prometheus scraping, and this already works however is it possible that the /metrics endpoint can be made public?

sagikazarmark:
This PR adds a brand new HTTP server to SFTPGo exposing the following endpoints:

- `/healthz`: for health checks
- `/metrics` for Prometheus metrics
- `/debug` for pprof

The code and configuration are identical to the same code found in the HTTP server, I just copied it over and simplified it a bit.

This PR does not address the following questions **on purpose**:

- what should happen to the same endpoints in the HTTP server? (IMHO they should be removed)
- potential refactoring of the configuration (eg. move the profiler enable flag under the telemetry config)
- should the telemetry server run for SFTP subsystem? (probably yes)

These questions should be addressed in separate PRs/issues.

This PR potentially fixes #244.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -16,6 +16,7 @@ import (
 	"github.com/drakkan/sftpgo/kms"
 	"github.com/drakkan/sftpgo/logger"
 	"github.com/drakkan/sftpgo/sftpd"
+	"github.com/drakkan/sftpgo/telemetry"
 	"github.com/drakkan/sftpgo/utils"
 	"github.com/drakkan/sftpgo/version"
 	"github.com/drakkan/sftpgo/webdavd"
@@ -38,14 +39,15 @@ var (
 )
 
 type globalConfig struct {
-	Common       common.Configuration  `json:"common" mapstructure:"common"`
-	SFTPD        sftpd.Configuration   `json:"sftpd" mapstructure:"sftpd"`
-	FTPD         ftpd.Configuration    `json:"ftpd" mapstructure:"ftpd"`
-	WebDAVD      webdavd.Configuration `json:"webdavd" mapstructure:"webdavd"`
-	ProviderConf dataprovider.Config   `json:"data_provider" mapstructure:"data_provider"`
-	HTTPDConfig  httpd.Conf            `json:"httpd" mapstructure:"httpd"`
-	HTTPConfig   httpclient.Config     `json:"http" mapstructure:"http"`
-	KMSConfig    kms.Configuration     `json:"kms" mapstructure:"kms"`
+	Common          common.Configuration  `json:"common" mapstructure:"common"`
+	SFTPD           sftpd.Configuration   `json:"sftpd" mapstructure:"sftpd"`
+	FTPD            ftpd.Configuration    `json:"ftpd" mapstructure:"ftpd"`
+	WebDAVD         webdavd.Configuration `json:"webdavd" mapstructure:"webdavd"`
+	ProviderConf    dataprovider.Config   `json:"data_provider" mapstructure:"data_provider"`
+	HTTPDConfig     httpd.Conf            `json:"httpd" mapstructure:"httpd"`
+	HTTPConfig      httpclient.Config     `json:"http" mapstructure:"http"`
+	KMSConfig       kms.Configuration     `json:"kms" mapstructure:"kms"`
+	TelemetryConfig telemetry.Conf        `json:"telemetry" mapstructure:"telemetry"`
 }
 
 func init() {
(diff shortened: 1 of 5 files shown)


drakkan:
> This PR adds a brand new HTTP server to SFTPGo exposing the following endpoints:
> 
> * `/healthz`: for health checks
> * `/metrics` for Prometheus metrics
> * `/debug` for pprof
> 

thank you for working on this!

> The code and configuration are identical to the same code found in the HTTP server, I just copied it over and simplified it a bit.
> 
> This PR does not address the following questions **on purpose**:
> 
> * what should happen to the same endpoints in the HTTP server? (IMHO they should be removed)

yes, it should be deprecated and removed after 1.3.x release

> * potential refactoring of the configuration (eg. move the profiler enable flag under the telemetry config)

good point, I think this will not break compatibility we should only document it

> * should the telemetry server run for SFTP subsystem? (probably yes)
> 

no, it should not run. In SFTP subsystem mode the main SSH server (OpenSSH) will start a new SFTPGo process for each SFTP session so we can have many SFTPGo instances running, we should not bind to any port in this mode

> These questions should be addressed in separate PRs/issues.
> 
> This PR potentially fixes #244.

sagikazarmark:
> no, it should not run. In SFTP subsystem mode the main SSH server (OpenSSH) will start a new SFTPGo process for each SFTP session so we can have many SFTPGo instances running, we should not bind to any port in this mode

Thanks for clarifying, I wasn't aware how sftp subsystem works.

sagikazarmark:
@drakkan when do you think 1.3 will be released?

The config file flag and the separate telemetry server would be great additions to the helm chart and would make a lot of things easier.

drakkan:
> @drakkan when do you think 1.3 will be released?
> 
> The config file flag and the separate telemetry server would be great additions to the helm chart and would make a lot of things easier.

I'm improving [FTP support](https://github.com/fclairamb/ftpserverlib/commits?author=drakkan) after finishing the work in ftpserverlib I have to integrate the changes in SFTPGo, this should be quite easy.

Currently we are using my pkg/sftp branch, see go.mod, I would like to use pkg/sftp again once [this PR](https://github.com/pkg/sftp/pull/397) is merged. Using pkg/sftp master branch the new sftpfs will not work and the `more-optimization` branch has currently some regressions. 

I also plan to add some other features, for example:

- a built-in defender to automatically blacklist clients that try too many authentications (based on a configuration that you can disable if you don't want it)
- improve quota update performance delaying the update. Currently if you upload 1000 small files SFTPGo will do a quota update for each file, we can accumulate the quota diff and update the quota, for example, once a minute
- different bandwidth limits based on the client IP
- various issues such as #253 and maybe #197

I'll plan to add SFTPv6 support to pkg/sftp too but I think I'll postpone this to 1.4.x, for 1.3.x I think I have to add statvfs support at least since I added the AVBL command to the FTP server.

I don't have a definite time frame, it depends on the time I can dedicate to the project, as a rough estimate I hope to release 1.3.0 before the end of January.

Doing small incremental release is time consuming (I have to manually build and test distro packages, prepare the release email etc...).

I think we can change the release model once all the major features are implemented, thank you for your patience

sagikazarmark:
> a built-in defender to automatically blacklist clients that try too many authentications (based on a configuration that you can disable if you don't want it)

That's definitely an interesting feature. I was looking at solutions like fail2ban, but sadly those don't really work with Kubernetes. Still, it feels like a waste of resources to implement this kind of security in all applications (again) when we (the IT world) already solved it beautifully with tools like fail2ban.

drakkan:
> > a built-in defender to automatically blacklist clients that try too many authentications (based on a configuration that you can disable if you don't want it)
> 
> That's definitely an interesting feature. I was looking at solutions like fail2ban, but sadly those don't really work with Kubernetes. Still, it feels like a waste of resources to implement this kind of security in all applications (again) when we (the IT world) already solved it beautifully with tools like fail2ban.

Currently you can use [post-connect](https://github.com/drakkan/sftpgo/blob/master/docs/post-connect-hook.md) and [post-login](https://github.com/drakkan/sftpgo/blob/master/docs/post-login-hook.md) hooks, some users implemented their protector in this way. 

I'm not aware of something like fail2ban that works in every use case, if there is such a solution I will gladly avoid this development :)

sagikazarmark:
fail2ban and similar tools scan logs of applications and set firewall rules to ban IP addresses on the network level. Unfortunately, that doesn't work on Kubernetes because there isn't really firewall involved (except on the cloud provider side, but that's a bit different story).

Currently I use IP address limitation, but that still reaches sftpgo itself. I'm currently looking at firewall settings to limit access to sftpgo on the network level, but an extra level of security never hurts.

Issue #225, opened 2020-11-17
web admin clone user (aka template)


fdefilippo:
Hi, do you think it is feasible to add the ability to clone a user to use as a template for creating new ones?

thnx

drakkan:
I'm not very fluent with web interfaces. It can be done if someone contributes it :)

PR #259, opened 2020-12-25
Feature: Clone existing user via web admin

zefixlluja:
UI based cloning of an existing user. The "add user" screen is prepopulated with existing user data.

Resolves drakkan/sftpgo#225


drakkan:
Hi,

thank you for this PR and sponsorship, both of which are greatly appreciated.

Regarding the PR it has some minor issues:

- if you clone a user with a kms secret, the secret will remain redacted and therefore it will not work for the new user, we must decrypt any KMS secret before cloning
- the index of the buttons for deleting and scanning quotas has changed since the clone button was added before them, we need to update some javascript

I'll merge the PR as is and I'll fix these minor issues in a successive commit, thank you!

Issue #275, opened 2021-01-12
S3: stat potentially paginates through thousands of objects to find zero-byte-keyed dirs


trapped:
For context, I'm using the S3 storage backend pointed at a custom S3-compatible server (Backblaze B2) with the SFTP frontend. The bucket contains millions of files amounting to tens of TBs of data.
Clients connect via SFTP, create a new directory and upload a few files in there.

`Mkdir()` at root level *consistently* times out (`error creating dir: "/testdir" error: RequestCanceled: request context canceled; caused by: context deadline exceeded`), looking for zero-byte-keyed dirs through `Stat()` first:

https://github.com/drakkan/sftpgo/blob/a8a17a223a0bc35941366c8b462fa7ceaff9bfdb/vfs/s3fs.go#L123-L131

it then calls `getStatCompat()`, which ends up paginating through **all** files in the bucket (if the directory doesn't yet exist):

https://github.com/drakkan/sftpgo/blob/a8a17a223a0bc35941366c8b462fa7ceaff9bfdb/vfs/s3fs.go#L139-L142

Looking at the code it seems like `getStatCompat()` is only used for finding the aforementioned zero-byte keyed objects; is there any reason not to simply `HEAD <name>/`, which correctly returns metadata for zero-byte keys?

I've implemented the change to `Stat()` in a local branch and it works fine; if there aren't any edge cases or incompatibilities that `getStatCompat()` already accounts for I'll happily submit it in a PR.

Thanks!

drakkan:
Hi,

`getStatCompat()` is there only for backward compatibility because initially we created "directories" as keys with a `/` suffix, we set the content type to `inode/directory` now, this method should be removed in future, for now a `HEAD <name>/` is ok, please submit a PR, thank you!

trapped:
Ah, I see then. Do you think we should explicitly check that `Content-Type` is `inode/directory` and `Content-Size` is `0` before treating the object as a directory?

Also, PR open: #276

drakkan:
> Ah, I see then. Do you think we should explicitly check that `Content-Type` is `inode/directory` and `Content-Size` is `0` before treating the object as a directory?
> 

we should do something like [this](https://github.com/drakkan/sftpgo/blob/master/vfs/gcsfs.go#L122)

> Also, PR open: #276

PR #276, opened 2021-01-12
S3fs: don't paginate to find zero-byte-keyed dirs

trapped:
This PR removes `getStatCompat()` and uses `HEAD <name>/` to stat "directories"; see #275

### DIFF
--- a/vfs/s3fs.go
+++ b/vfs/s3fs.go
@@ -139,7 +139,10 @@ func (fs *S3Fs) Stat(name string) (os.FileInfo, error) {
 	// with a forwarding slash.
 	obj, err = fs.headObject(strings.TrimPrefix(name, "/") + "/")
 	if err == nil {
-		return NewFileInfo(name, true, 0, time.Now(), false), nil
+		objSize := *obj.ContentLength
+		objectModTime := *obj.LastModified
+		isDir := *obj.ContentType == dirMimeType
+		return NewFileInfo(name, isDir, objSize, objectModTime, false), nil
 	}
 	if !fs.IsNotExist(err) {
 		return result, err
@@ -584,19 +587,6 @@ func (fs *S3Fs) resolve(name *string, prefix string) (string, bool) {
 		result = result[:i]
 	}
 	return result, isDir
-}
-
-func (fs *S3Fs) isEqual(s3Key *string, virtualName string) bool {
-	if *s3Key == virtualName {
-		return true
-	}
-	if "/"+*s3Key == virtualName {
-		return true
-	}
-	if "/"+*s3Key == virtualName+"/" {
-		return true
-	}
-	return false
 }
 
 func (fs *S3Fs) checkIfBucketExists() error {
--- a/vfs/s3fs.go
+++ b/vfs/s3fs.go
@@ -138,51 +138,14 @@ func (fs *S3Fs) Stat(name string) (os.FileInfo, error) {
 	}
 	// the requested file could still be a directory as a zero bytes key
 	// with a forwarding slash.
-	// As last resort we do a list dir to find it
-	return fs.getStatCompat(name)
-}
-
-func (fs *S3Fs) getStatCompat(name string) (os.FileInfo, error) {
-	var result FileInfo
-	prefix := path.Dir(name)
-	if prefix == "/" || prefix == "." {
-		prefix = ""
-	} else {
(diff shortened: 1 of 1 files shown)


vfs/s3fs.go
```diff
@@ -138,51 +137,14 @@ func (fs *S3Fs) Stat(name string) (os.FileInfo, error) {
...
-	var result FileInfo
-	prefix := path.Dir(name)
-	if prefix == "/" || prefix == "." {
-		prefix = ""
-	} else {
-		prefix = strings.TrimPrefix(prefix, "/")
-		if !strings.HasSuffix(prefix, "/") {
-			prefix += "/"
-		}
+	obj, err = fs.headObject(strings.TrimPrefix(name, "/") + "/")
```
drakkan, line 140: is TrimPrefix required here? We do `obj, err := fs.headObject(name)` some lines above `obj, err := fs.headObject(name+"/")` should work

vfs/s3fs.go
```diff
@@ -138,51 +137,14 @@ func (fs *S3Fs) Stat(name string) (os.FileInfo, error) {
...
+	if err == nil {
+		return NewFileInfo(name, true, 0, time.Now(), false), nil
```
drakkan, line 142: we could do something like this here:

```
               objSize := *obj.ContentLength
		objectModTime := *obj.LastModified
		return NewFileInfo(name, true, objSize, objectModTime, false), nil
```

vfs/s3fs.go
```diff
@@ -138,51 +137,17 @@ func (fs *S3Fs) Stat(name string) (os.FileInfo, error) {
...
-	} else {
-		prefix = strings.TrimPrefix(prefix, "/")
-		if !strings.HasSuffix(prefix, "/") {
-			prefix += "/"
-		}
+	obj, err = fs.headObject(strings.TrimPrefix(name, "/") + "/")
+	if err == nil {
+		objSize := *obj.ContentLength
+		objectModTime := *obj.LastModified
+		isDir := *obj.ContentType == dirMimeType
```
drakkan, line 144: if we get here we created a dir with an old SFTPGo version or using s3fs and the content type is not dirMimeType, we should add this check for the first stat, here we should always set isDir to true, sorry if I was not clear enough

vfs/s3fs.go
drakkan: I'll remove the TrimPrefix if it is not required, thank you

trapped: Testing right now (through the Minio Python client, connecting to B2 via its S3 API):
```py
import minio
c = minio.Minio(...)
# create a 0-bytes-keyed empty obj "testdir/"
c.put_object(
    bucket_name='test',
    object_name='testdir/',
    data=io.BytesIO(b''),
    length=0,
    content_type='inode/directory') # 'd41d8cd98f00b204e9800998ecf8427e'
# HEAD /<name>
c.stat_object('test', '/testdir') # NoSuchKey: message: The specified key does not exist.
# HEAD <name>
c.stat_object('test', 'testdir') # NoSuchKey: message: The specified key does not exist.
# HEAD <name>/
c.stat_object('test', 'testdir/') # <minio.definitions.Object at 0x7fea85112b90>
# cleanup
c.remove_object('test', 'testdir/')
# create a 0-bytes-keyed empty obj "/testdir/"
c.put_object(
    bucket_name='test',
    object_name='/testdir/',
    data=io.BytesIO(b''),
    length=0,
    content_type='inode/directory') # 'd41d8cd98f00b204e9800998ecf8427e'
# HEAD /<name>/
c.stat_object('test', '/testdir/') # <minio.definitions.Object at 0x7fea85112b90>
# HEAD /<name>
c.stat_object('test', '/testdir') # NoSuchKey: message: The specified key does not exist.
# HEAD <name>
c.stat_object('test', 'testdir') # NoSuchKey: message: The specified key does not exist.
# HEAD <name>/
c.stat_object('test', 'testdir/') # NoSuchKey: message: The specified key does not exist.
```
Both `/<name>/` and `name/` are valid, but at root `/<name>/` is considered a file rather than a directory, at least by the Backblaze web UI:

2021-01-12-174804_702x160_scrot

Also, testing with directories created by sftpgo itself it seems they are created as `<name>/` rather than `/<name>/`, while `name` in this case starts with `/` (that's why I used `strings.TrimPrefix`).

trapped:
Closing in favor of #277

PR #277, opened 2021-01-12
s3: don't paginate to find zero-byte-keyed dirs

### LINKED ISSUE #275: S3: stat potentially paginates through thousands of objects to find zero-byte-keyed dirs

drakkan:
Fixes #275

@trapped can you please try this patch with your S3 compatible backed?  The initial '/' prefix should be automatically removed when SFTPGo create/check a key.

Of course I gave you the attribution for this patch.

I was a bit confused sorry, for GCS and AzBlob backeds we can get the content-type when we list the keys, for S3 this is not possibile so we have to add the `/` suffix when creating directories so this is not compact code that we could remove. I added some comments to clarify this. Thank you for spotting this issue.

I pushed a similar fix for GCS, the main difference is that for GCS we can remove the check in future since we don't create anymore (since 1.2.x) directories with a trailing `/`.

### DIFF
--- a/vfs/s3fs.go
+++ b/vfs/s3fs.go
@@ -4,7 +4,6 @@ package vfs
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"mime"
 	"net/url"
@@ -122,6 +121,7 @@ func (fs *S3Fs) Stat(name string) (os.FileInfo, error) {
 	}
 	obj, err := fs.headObject(name)
 	if err == nil {
+		// a "dir" has a trailing "/" so we cannot have a directory here
 		objSize := *obj.ContentLength
 		objectModTime := *obj.LastModified
 		return NewFileInfo(name, false, objSize, objectModTime, false), nil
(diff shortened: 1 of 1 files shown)


trapped:
@drakkan just tested, seems to work well!

drakkan:
> @drakkan just tested, seems to work well!

Thank you and sorry for the confusing info. I think we should also disallow files with a trailing `/` I'll test and eventually fix this after the ongoing API refactoring

trapped:
> Thank you and sorry for the confusing info. I think we should also disallow files with a trailing `/` I'll test and eventually fix this after the ongoing API refactoring

Thank you for the amazing work!

drakkan:
if you use SFTPGo at your company please consider a [sponsorship](https://github.com/sponsors/drakkan) :) thank you!

PR #305, opened 2021-02-07
feat(docker): add permissions to data dirs

sagikazarmark:
This change should improve permission handling for mounted volumes in a running container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,12 +32,14 @@ RUN apt-get update && apt-get install --no-install-recommends -y ca-certificates
 
 RUN if [ "${INSTALL_OPTIONAL_PACKAGES}" = "true" ]; then apt-get update && apt-get install --no-install-recommends -y git rsync && rm -rf /var/lib/apt/lists/*; fi
 
-RUN mkdir -p /etc/sftpgo /var/lib/sftpgo /usr/share/sftpgo /srv/sftpgo
+RUN mkdir -p /etc/sftpgo /var/lib/sftpgo /usr/share/sftpgo /srv/sftpgo/{data,backups}
 
 RUN groupadd --system -g 1000 sftpgo && \
     useradd --system --gid sftpgo --no-create-home \
     --home-dir /var/lib/sftpgo --shell /usr/sbin/nologin \
     --comment "SFTPGo user" --uid 1000 sftpgo
+
+RUN chown -R 1000:1000 /srv/sftpgo
 
 COPY --from=builder /workspace/sftpgo.json /etc/sftpgo/sftpgo.json
 COPY --from=builder /workspace/templates /usr/share/sftpgo/templates
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -39,10 +39,12 @@ RUN if [ "${INSTALL_OPTIONAL_PACKAGES}" = "true" ]; then apk add --update --no-c
 # https://github.com/gliderlabs/docker-alpine/issues/367#issuecomment-424546457
 RUN test ! -e /etc/nsswitch.conf && echo 'hosts: files dns' > /etc/nsswitch.conf
 
-RUN mkdir -p /etc/sftpgo /var/lib/sftpgo /usr/share/sftpgo /srv/sftpgo
+RUN mkdir -p /etc/sftpgo /var/lib/sftpgo /usr/share/sftpgo /srv/sftpgo/{data,backups}
 
 RUN addgroup -g 1000 -S sftpgo && \
     adduser -u 1000 -h /var/lib/sftpgo -s /sbin/nologin -G sftpgo -S -D -H -g "SFTPGo user" sftpgo
+
+RUN chown -R 1000:1000 /srv/sftpgo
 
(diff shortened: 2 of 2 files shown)


drakkan:
Hi,

thank you for this PR, I'll do more tests later today, I just did a quick test. We don't have bash inside the runtime stage so with this change we will have:

```
$ ls
{data,backups}
```

we also already have a chown command so we could change it instead of adding a new RUN command.

I suggest to change the PR like this:

```
diff --git a/Dockerfile b/Dockerfile
index 712fe6c..3b98b20 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y ca-certificates
 
 RUN if [ "${INSTALL_OPTIONAL_PACKAGES}" = "true" ]; then apt-get update && apt-get install --no-install-recommends -y git rsync && rm -rf /var/lib/apt/lists/*; fi
 
-RUN mkdir -p /etc/sftpgo /var/lib/sftpgo /usr/share/sftpgo /srv/sftpgo
+RUN mkdir -p /etc/sftpgo /var/lib/sftpgo /usr/share/sftpgo /srv/sftpgo/data /srv/sftpgo/backups
 
 RUN groupadd --system -g 1000 sftpgo && \
     useradd --system --gid sftpgo --no-create-home \
@@ -55,9 +55,7 @@ RUN sed -i "s|\"users_base_dir\": \"\",|\"users_base_dir\": \"/srv/sftpgo/data\"
     sed -i "s|\"backups\"|\"/srv/sftpgo/backups\"|" /etc/sftpgo/sftpgo.json && \
     sed -i "s|\"address\": \"127.0.0.1\",|\"address\": \"\",|" /etc/sftpgo/sftpgo.json
 
-COPY ./docker/scripts/entrypoint.sh /docker-entrypoint.sh
-
-RUN chown -R sftpgo:sftpgo /etc/sftpgo && chown sftpgo:sftpgo /var/lib/sftpgo /srv/sftpgo
+RUN chown -R sftpgo:sftpgo /etc/sftpgo /srv/sftpgo && chown sftpgo:sftpgo /var/lib/sftpgo
 
 WORKDIR /var/lib/sftpgo
 USER 1000:1000
diff --git a/Dockerfile.alpine b/Dockerfile.alpine
index 8f72bec..3fcb815 100644
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -39,7 +39,7 @@ RUN if [ "${INSTALL_OPTIONAL_PACKAGES}" = "true" ]; then apk add --update --no-c
 # https://github.com/gliderlabs/docker-alpine/issues/367#issuecomment-424546457
 RUN test ! -e /etc/nsswitch.conf && echo 'hosts: files dns' > /etc/nsswitch.conf
 
-RUN mkdir -p /etc/sftpgo /var/lib/sftpgo /usr/share/sftpgo /srv/sftpgo
+RUN mkdir -p /etc/sftpgo /var/lib/sftpgo /usr/share/sftpgo /srv/sftpgo/data /srv/sftpgo/backups
 
 RUN addgroup -g 1000 -S sftpgo && \
     adduser -u 1000 -h /var/lib/sftpgo -s /sbin/nologin -G sftpgo -S -D -H -g "SFTPGo user" sftpgo
@@ -60,7 +60,7 @@ RUN sed -i "s|\"users_base_dir\": \"\",|\"users_base_dir\": \"/srv/sftpgo/data\"
     sed -i "s|\"backups\"|\"/srv/sftpgo/backups\"|" /etc/sftpgo/sftpgo.json && \
     sed -i "s|\"address\": \"127.0.0.1\",|\"address\": \"\",|" /etc/sftpgo/sftpgo.json
 
-RUN chown -R sftpgo:sftpgo /etc/sftpgo && chown sftpgo:sftpgo /var/lib/sftpgo /srv/sftpgo
+RUN chown -R sftpgo:sftpgo /etc/sftpgo /srv/sftpgo && chown sftpgo:sftpgo /var/lib/sftpgo
 
 WORKDIR /var/lib/sftpgo
 USER 1000:1000
```

These directories will be created with the right permissions as needed, do they need to be pre-created to enforce some security policy that can't be applied if the directory doesn't exist?

If you mount a volume on `/srv/sftpgo`, as documented, these directories may be missing on the volume itself, can this be problematic?

thank you!

P.S. The backup dir, if auto-created, will have 0700 as permissions, maybe we can do the same here.

sagikazarmark:
> These directories will be created with the right permissions as needed, do they need to be pre-created to enforce some security policy that can't be applied if the directory doesn't exist?

The idea is that data and backups might be different mounts, so precreating the directories might be better.

drakkan:
> > These directories will be created with the right permissions as needed, do they need to be pre-created to enforce some security policy that can't be applied if the directory doesn't exist?
> 
> The idea is that data and backups might be different mounts, so precreating the directories might be better.

ok I like the idea, we should to add a documentation note too.

drakkan:
I implemented this feature myself, thank your for the idea!

sagikazarmark:
Oh, sorry I couldn't finish it.

drakkan:
> Oh, sorry I couldn't finish it.

No problem, it should work fine, please let me know if you find any issue

PR #335, opened 2021-03-04
Add in-memory filesystem implementation

hslatman:
The in-memory filesystem implementation is backed by
spf13/afero. This resulted in a fairly shallow layer
around the methods provided by afero.

So far the basic SFTP commands work, but there are
some fixes to be made, for example around symlinks and
determining paths. These probably also need some tests.

### DIFF
--- a/dataprovider/dataprovider.go
+++ b/dataprovider/dataprovider.go
@@ -1396,6 +1396,13 @@ func validateFilesystemConfig(user *User) error {
 		user.FsConfig.AzBlobConfig = vfs.AzBlobFsConfig{}
 		user.FsConfig.CryptConfig = vfs.CryptFsConfig{}
 		return nil
+	} else if user.FsConfig.Provider == MemoryFilesystemProvider {
+		// TODO: validation
+		user.FsConfig.S3Config = vfs.S3FsConfig{}
+		user.FsConfig.GCSConfig = vfs.GCSFsConfig{}
+		user.FsConfig.AzBlobConfig = vfs.AzBlobFsConfig{}
+		user.FsConfig.CryptConfig = vfs.CryptFsConfig{}
+		return nil
 	}
 	user.FsConfig.Provider = LocalFilesystemProvider
 	user.FsConfig.S3Config = vfs.S3FsConfig{}
--- a/dataprovider/user.go
+++ b/dataprovider/user.go
@@ -172,6 +172,7 @@ const (
 	AzureBlobFilesystemProvider                           // Azure Blob Storage
 	CryptedFilesystemProvider                             // Local encrypted
 	SFTPFilesystemProvider                                // SFTP
+	MemoryFilesystemProvider                              // Memory
 )
 
 // Filesystem defines cloud storage filesystem details
@@ -255,6 +256,8 @@ func (u *User) GetFilesystem(connectionID string) (vfs.Fs, error) {
 		return vfs.NewCryptFs(connectionID, u.GetHomeDir(), u.FsConfig.CryptConfig)
 	case SFTPFilesystemProvider:
 		return vfs.NewSFTPFs(connectionID, u.FsConfig.SFTPConfig)
+	case MemoryFilesystemProvider:
+		return vfs.NewMemFs(connectionID, u.GetHomeDir(), u.VirtualFolders)
 	default:
 		return vfs.NewOsFs(connectionID, u.GetHomeDir(), u.VirtualFolders), nil
 	}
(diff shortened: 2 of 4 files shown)


hslatman:
Sprinkled some prints for debugging; will have to clean that up.

drakkan:
Hi,

thank you for your contribution, but can you please explain your use case? SFTPGo is an application, not a library, why should we support a mem vfs? If this is not for testing, a user could upload a file bigger than the available memory: isn't a tmpfs better here? It has a size limit and you can use the standard os fs

hslatman:
Hi @drakkan,

Let me start by thanking you for creating SFTPGo. It's a great piece of useful software 😃 

My use case is indeed for testing and development purposes. Integration testing, specifically. Some of my work involves SFTP file transfers. Currently my options for testing these are broadly 1) to provide an alternative filesystem in the application itself for dev/test purposes (injected as dependency or some other method for abstracting it away), 2) temporarily set up a full SFTP server or 3) simply don't test the SFTP process itself. 

Ideally I would like to test such a process with a full functioning SFTP server that is easy to configure and setup/teardown (for which SFTPGo is clearly a good candidate), so that the SFTP specific functionality can also be tested (e.g. are private keys configured and used correctly in the code, are network related errors handled correctly, etc.). The fact that it's a full functioning SFTP server also helps in debugging during development, because the state of the server and files stored can be inspected manually using any client that supports SFTP.

I share your concern regarding the maximum memory size. I think it's possible to largely mitigate the issue by providing a conservative default maximum amount of memory for storing files (i.e. like X% of total system memory) by implementing `GetAvailableDiskSize()`, but correct me if I'm wrong. I could also add additional configuration for the maximum amount of memory to use. In the end this can still result in too many (big) files being sent to memory, but I think that can be explained and understood by the user.

Providing a `tmpfs` implementation is also an option; either as a replacement or complementary to the in-memory one, if you ask me. Your points are spot on; probably less issues with the maximum size out of the box, but as mentioned before I think those can be tackled adequately. I think I can use `afero` with its `TempDir` function for this. 

An, IMO, nice benefit of having an option to run SFTPGo with a memfs (or tmpfs) storage backend for development and tests, is that you can use the same technology in production, reducing operational complexity overall.

drakkan:
Hi,

I think for testing purpose is better to use [pkg/sftp](https://github.com/pkg/sftp) (I'm a contributor there and `pkg/sftp` is used as base for the SFTP implementation within SFTPGo). `pkg/sftp` already has a memfs implementation exactly for testing purpose and it try to support all the features you find in local fs (symlinks etc..). 

You can find [here](https://github.com/pkg/sftp/blob/master/examples/request-server/main.go) a sample SFTP server that use a memfs.

The `GetAvailableDiskSize` is only a method used to get the response for the `StatVFS` SFTP method, most clients don't use it before uploading a file and even if used it is only informational it don't limit anything.

I'm not sure that adding this implementation in SFTPGo is a good idea, as said this is an application and not a library and this implementation will basically duplicate something that `pkg/sftp` already provide. Anyway let me some time more to think about it. Thank you for your understanding

hslatman:
Hi @drakkan,

Thank you for pointing me to the `InMemHandler()`; I think that may be of use to me when testing in Go applications.

In most of my current work the application itself is not written in Go, however, so I'll have to talk the SFTP protocol. Having a quick to setup/teardown SFTP server, either manually or in Docker, without having to think about file persistence, would be great for this use case. The additional benefit of when it's available in SFTPGo is that there's only a single piece of technology in either of our development, test and production environments. 

Bit of a bummer that the limit is not used to determine if the disk is full, but that's what it is. Maybe I can find another way to do it 🙂. 

So, if after you consider the above, still think that `memfs` is not the way to go, would you suggest that a new `tmpfs` implementation would have a bigger chance to get added? I think a `tmpfs` has similar benefits as the `memfs` one, can be implemented similarly as the `memfs` one and I probably can make some time to do it.

drakkan:
With docker you can simply use something like this:

```
docker run --rm --name some-sftpgo -p 8080:8080 -p 2022:2022 drakkan/sftpgo:edge-slim
```

and you won't have file persistence while you can still use a real fs implementation with no limitations.

For `tmpfs` I meant a real tmpfs something like this:

```
mount | grep /tmp
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,nr_inodes=409600,inode64)
```

and so external to SFTPGo

hslatman:
Both are fair solutions and they'll probably be used in practice, but require a bit more infra and/or know-how. It may be an idea to put these instructions in a document/wiki, perhaps?

In my opinion having a `tmpfs` implementation built into the application itself seems slightly more convenient, but that's more of a personal preference than anything else.

drakkan:
Hi,

the implementation here is a draft, please use it in your separate branch if you prefer this solution to the docker based one. Once it is more mature, point me to your branch or submit a new PR or reopen this one.

To be useful I think it should be configurable to emulate local filesystem and/or the supported cloud backends. If instead this is an implementation with its own behavior and different from the others I think it is not so useful. If you test something using the memfs and then get a different behavior with the local filesystem, I don't see how this can be useful. 
If instead you can configure memfs to emulate the S3 implementation and thus you can test S3 without configuring MinIO or any other backend, this will become interesting.

I'll document the docker based solution for now, thank you

PR #355, opened 2021-03-22
Added in a first draft of the page related to sftpgo-ldap-http-server.

orware:
Hi Nicola,

I apologize, I ended up not having as much time this weekend as I wanted to for writing this, so I'm not sure if this is where you'd like things to be as a starting point for inclusion in the examples area, but we can adjust if needed. The Youtube video was only uploaded a short while ago so it's higher quality versions are still being processed at the moment.

(I wasn't really sure on the folder name to use for this PR, so I ended up using one with "php" and "activedirectory" in the folder name which might be a little more descriptive, since I felt `sftpgo-ldap-http-server` might not make as much sense for others if they saw that listed in the examples folder, but I'm welcome to making any adjustments you feel would be helpful.

Thank you and I'm hoping this will help others in getting started with SFTPGo!


drakkan:
Thank you @orware , it looks perfect! I would add a link to the video in the howto section too if you agree, but this could be done also in a separate commit

Issue #153, opened 2020-08-09
S3 Versioning


fdefilippo:
How about supporting s3 buckets with versioning enabled? Minio with the new API 7.x supports them.
thnx

<pre>
$ s3mc ls --versions local/test
[2020-07-11 10:39:00 CEST]  1.7KiB group (v3, 14d06c4b-a2a0-4ab7-afe1-e1edb8cd8676)
[2020-07-11 10:38:59 CEST]  1.7KiB group (v2, 6f02c9fd-da14-497c-8355-e0350a275f4e)
[2020-07-11 10:38:58 CEST]  1.7KiB group (v1, 3abd225c-ad0e-4f32-934b-806823053959)
[2020-07-23 23:06:38 CEST]    615B send.sh (v2, e9b064e3-a5ad-4756-b94f-b57ba48b9853)
[2020-07-23 23:06:35 CEST]    615B send.sh (v1, 2d1acf6e-ff5b-43e9-95f3-0dd5f1dcf02e)
[2020-06-02 16:54:02 CEST]  1.1KiB test.sh (v1, null)
[2020-07-22 12:13:12 CEST]      0B uno (v6, 4827a02d-4860-4ae9-bd79-b33299330317, delete-marker)
[2020-07-12 11:40:18 CEST]  1.7KiB uno (v5, 2da7024f-e7f9-4ba0-9810-66142b881016)
[2020-07-12 11:39:48 CEST]  1.7KiB uno (v4, c63802ef-28a7-4216-bb61-05976bbb52cc)
[2020-07-12 11:30:59 CEST]  1.7KiB uno (v3, 7bad9ec9-7bce-4dce-bdbf-caddfef44896)
[2020-07-12 11:30:27 CEST]  1.7KiB uno (v2, 77f02b51-f09d-4273-9d9c-5c842008a123)
[2020-07-12 11:30:12 CEST]  1.7KiB uno (v1, 28c2e63a-9d6c-432f-98b8-6748deba9d52)
</pre>

drakkan:
Hi,

the next week I should have access to a versioned S3 bucket (thank to one of the SFTPGo sponsors) and I'll evaluate this feature in more detail. 

My first impression is that you can enable versioning and recover files using the AWS console, while you should see the current version via SFTP/FTP. Basically I agree with [s3fs maintainers](https://github.com/s3fs-fuse/s3fs-fuse/issues/67). 

[...]

PR #449, opened 2021-06-06
move Filesystem config validation to vfs

mreithub:
Hello sftpgo team,

since I'm new to this project and feel a bit like barging in the front door with a potentially huge changeset I first wanted to let you know that I'm fully aware that this - right now - is only an idea in my head and that it may not be the direction you want the project to move in (sometimes I can't help myself and just need to refactor stuff...)

I've tried to split things up into different PRs to make the changes easier to verify, but this also means it may take some time for you to see the what I'm trying to achieve...  

This is the first PR in a series. so: don't merge this PR before you've seen a few of the others and please feel free to ask questions or provide input to steer me in a specific direction. But please let me know what you think :)

I'm trying to improve vfs modularity. Right now there are `switch fsConfig.Provider {...}` statements pretty much all around the code.  
The first step is to simply move that code into the vfs package and into each of the different vfs implementations in a second one.  

The goal I'm aiming for is to have each vfs implementation register itself, allowing us to eventually get rid of the `vfs/*_disabled.go` files (and potentially allowing niche-vfs implementations to live in separate forks without conflicting with each other)

Let me know if I should do things differently (or if this isn't something you'll likely merge)

oh god, this PR is getting longer and longer, sorry for that... :facepalm: 

but let's get started :)

----

This PR is the first in a series to move `vfs.FilesystemProvider`-specific code into the `vfs` package.

It only moves vfs-specific code blocks from various places in the code into the vfs package. in a later step this code will be further split up into individual methods for each `vfs.*Config` struct.

This PR:
- only changes code (all the config files, JSON schemas and HTTP APIs remain the same)
- moves `dataprovider.validateFilesystemConfig()` to `vfs.Filesystem.ValidateConfig()`
- added `vfs.Filesystem.HasRedactedSecrets()` and moved code from `dataprovider.User.HasRedactedSecrets()` there
- moved `dataprovider.ValidationError` to package `vfs`, renamed  `dataprovider.vfsValidatorHelper` to `vfs.ValidatorHelper`
  I needed to use them in the vfs package and couldn't do that if they remained in the dataprovider package (import cycles)


mreithub:
I'll add a quick comment here to tell you a bit more about my use case (may help you understand my intentions here a bit better):

I'm looking for a way to have users upload files (e.g. using filezilla) which then automatically get processed (extracting metadata, converting to different formats, uploading to s3, ...)

Right now they upload to a classic sftp server's local filesystem and a script of mine traverses the whole tree once an hour looking for new files and processing them. I want to replace that local fs directory with S3 (for various reasons) but want to avoid naively listing hundreds of thousands files multiple times an hour when >99% of them likely haven't changed.

What I've toyed around with in sftpgo was writing a wrapper to s3fs that keeps the directory tree in a Postgres DB - which gives me the ability to quickly query for recently changed files (allowing me to run the post-processing script more efficiently and therefore a lot more often).

I've also played around with implicit versioning - the directory tree in psql has start and end timestamp fields and a content_hash for each file - and the s3 objects use that content_hash as filename.
I know S3 has versioning support - but not all s3-compatible services do and to be honest I haven't checked `vfs/s3fs.go` to see if it implements it.
Since I need/want to do some sort of versioning in the directory tree stored in psql anyway, I pretty much get versioned file contents for free - this also gives us file-level deduplication for free.
But using the content hash as object name means the S3 bucket is pretty much unusable without the directory tree stored in psql - which is fine for our use case but  drastically reduces the number of potential other users -> reducing the value of merging it into the official project.

I'm trying to write it in a way that may be useful to others aswell and am open to eventually include it in sftpgo but I kinda expect that it's a bit out-of-scope for the main project.

I'm not sure we'll end up using this vfs wrapper in the way I've described it here, but I'd like to know that it will be maintainable even outside the main sftpgo project.

(I accidentally nuked that code while trying to keep these pull requests clean but I'll upload it to github once it's somewhat usable.)

drakkan:
Hi,

I like this contribution, thank you! After the release of v2.1.0 (which should happen soon) I plan to add support for a plugin system, probably based on something like [this](https://github.com/hashicorp/go-plugin) and your proposed change could fit perfectly.

Regarding S3 versioning, SFTPGo does not support it, I think we need a database backend even with S3 versioning, see #153, I'm not sure if this will ever happen in the main tree (it could be hard to support it with a provider like bolt for example), maybe using a plugin.

Instead of scanning the filesystem for changes at regular intervals, have you considered using SFTPGo's [hook system](https://github.com/drakkan/sftpgo/blob/main/docs/custom-actions.md)? You can get an HTTP notification (or run an external command) as soon as a new file is uploaded, and you can apply your logic there.

Thank you!

mreithub:
Thanks for pointing me to the right page about hooks. I thought about the possibility of using them but I must have overlooked that document (I only found the ones for login, checkPassword, ...)

I thought about using hooks, but I'm slightly concerned that something may break silently in the long run (distributed systems are hard).
If I were to go that route, it would probably be a combination of the pre- and post-upload hooks (pre-upload to create a stub in the database so I could better track missing or failed post-upload hook runs, post-upload doing the actual processing).

I haven't looked too deep into 

Another reason why I was leaning towards a vfs wrapper was that I kind of assumed s3 directory listings would be a lot slower (request lag + overhead when listing thousands of files at once).
I've had that experience in the past (I think while experimenting with s3fs or goofys or something like that), but trying it out over the weekend the response times with sftpgo+s3+filezilla seemed fine).
Yet another thing I thought about was the ability to add other features s3 lacks (e.g. file permissions, the ability to set mtimes, etc. - a lot of our old code still uses filesize+mtime to decide whether a file has been changed - as does filezilla in some of the upload modes)

I've seen that sftpgo seems to provide its own permission scheme but I haven't looked that deep into it either yet :)
(the whole virtual folder thing is also still a bit of a mystery to me)

Anyway, I'm still in the process of evaluating these options...

drakkan:
> Thanks for pointing me to the right page about hooks. I thought about the possibility of using them but I must have overlooked that document (I only found the ones for login, checkPassword, ...)
> 
> I thought about using hooks, but I'm slightly concerned that something may break silently in the long run (distributed systems are hard).

If you use HTTP notifications, SFTPGo will retry them, on error, a configurable number of times (with a configurable backoff) before giving up

> If I were to go that route, it would probably be a combination of the pre- and post-upload hooks (pre-upload to create a stub in the database so I could better track missing or failed post-upload hook runs, post-upload doing the actual processing).
> 

pre- * hooks are also there for this purpose. Also, if your system is down, the pre-upload hook can forbid new uploads until your system is up again

> I haven't looked too deep into
> 
> Another reason why I was leaning towards a vfs wrapper was that I kind of assumed s3 directory listings would be a lot slower (request lag + overhead when listing thousands of files at once).

SFTPGo uses the AWS SDK directly, without any wrapper, to have more granular control on how to interact with S3, GCS, AzBlob, however we have to do HTTP calls, they are slower than direct filesystem calls for obviuos reasons :)

> I've had that experience in the past (I think while experimenting with s3fs or goofys or something like that), but trying it out over the weekend the response times with sftpgo+s3+filezilla seemed fine).
> Yet another thing I thought about was the ability to add other features s3 lacks (e.g. file permissions, the ability to set mtimes, etc. - a lot of our old code still uses filesize+mtime to decide whether a file has been changed - as does filezilla in some of the upload modes)
> 

reading mtime and file size is supported, the described usage pattern should already work. S3 does not allow to change the modification time directly, to support chtimes we should add custom metadata, but I think we need to perform an head request to retrieve them. This will be a big performance penality for directory listing. I'm not going to add support for this, so you need a vfs wrapper if you have this requirement. 

> I've seen that sftpgo seems to provide its own permission scheme but I haven't looked that deep into it either yet :)

I suggest using the SFTPGo permissions abstraction, they work the same regardless of the OS and cloud backend

> (the whole virtual folder thing is also still a bit of a mystery to me)
> 

take a look [here](https://github.com/drakkan/sftpgo/blob/main/docs/virtual-folders.md). The [getting started guide](https://github.com/drakkan/sftpgo/blob/main/docs/howto/getting-started.md) could help too

> Anyway, I'm still in the process of evaluating these options...

I like both this patch and the other one, I could merge them as soon as I find some time to do a proper review, thanks

mreithub:
Thanks so much for all the info - but please don't feel obligated to go into that much detail :)

I've expected hooks to pretty much work exactly like that - them being somewhat asynchronous still means I'll have to write them in a way that avoids related issues. I've spent some more time thinking about that issue today and right now I'm leaning more towards the hook-implementation :)

> SFTPGo uses the AWS SDK directly, without any wrapper, to have more granular control on how to interact with S3, GCS, AzBlob, however we have to do HTTP calls, they are slower than direct filesystem calls for obviuos reasons :)

I was talking about the s3 itself being noticeable slower. I've ran into the same thing in the past - first using minio's s3 implementation in go which if I remember correctly behaves almost like local files. But I think it implicitly did extra roundtrips we had no control over (one was to get the object size if I remember correctly).

Back then the s3 bucket was in another datacenter (not available in the local one yet) which resulted in quite the performance hit.
Using minio's s3 client meant I could pretty much use it as a drop-in replacement for the local filesystem (which was actually a network mount), but that was an issue I could only fix by using the official s3 lib directly.

> .. S3 does not allow to change the modification time directly, to support chtimes we should add custom metadata, but I think we need to perform an head request to retrieve them. This will be a big performance penality for directory listing. I'm not going to add support for this, so you need a vfs wrapper if you have this requirement (now that I think about it that may also have had something to do with expensive roundtrips due to s3 ListObject paging...

I'm aware of that (I've studied the S3 API docs while evaluating s3fs and its alternatives in the past - which was one of the reasons I was thinking about implementing that stuff in an extra abstraction layer

> I suggest using the SFTPGo permissions abstraction, they work the same regardless of the OS and cloud backend

will do

> take a look [here](https://github.com/drakkan/sftpgo/blob/main/docs/virtual-folders.md). The [getting started guide](https://github.com/drakkan/sftpgo/blob/main/docs/howto/getting-started.md) could help too

Thanks, I'll have a look :)

> I like both this patch and the other one, I could merge them as soon as I find some time to do a proper review, thanks

I have a few other small patch ideas lined up but haven't had time to write them yet...

drakkan:
Hi,

I plan to merge your PRs after v2.1.0 that should happen the next weekend, no other major/breaking changes are planned, when you have some free time please rebase, thank you

mreithub:
Sure, just let me know when it's safe to do so (I'm busy working on the hook app right now, I might take 1 or 2 days to actually get to do it :) )

drakkan:
It should be safe to rebase now, I plan to push only bug fixes and minor improvements before tagging v2.1.0

drakkan:
merged with some minor changes

https://github.com/drakkan/sftpgo/commit/f19937b715d5cddf28222981c68fab186d9976e5

thank you

mreithub:
just had a look at the commit. It makes a lot more sense to have ValidationError in utils...
I originally wanted to move it to common (for some reason didn't see the utils package...), but ended up with cyclic imports.

I'll look at rebasing the other merge request today :)
(been a bit busy this week...)

drakkan:
Thank you, I'll take a look at the other PR later today, I you don't have time to rebase it I'll do it myself and you will keep the commit attribution as for this PR.

I think these changes are required to expose filesystems as plugins, that is something that I want to try to add for v2.2

Thank you

PR #520, opened 2021-08-30
updated username regex (and error messages) to allow backslashes

JohnTheNerd:
backslash characters, although admittedly sound suboptimal, actually work perfectly fine as usernames within SFTPGo. in fact, you can create users with backslashes as part of their names directly in the database even without this PR, but you can no longer change any settings in the user afterwards because it fails the regex validation.

this PR merely adds the backslash character to the regex and error messages. there are four backslashes in the regex because both regular expressions _and_ Go strings require backslash characters to be escaped.

tested:

- SFTP with OpenSSH's `sftp` command line tool
- WebDAV with WinSCP
- FTP with both the Linux command line tool, Windows Explorer, and WinSCP for good measure

### DIFF
--- a/dataprovider/admin.go
+++ b/dataprovider/admin.go
@@ -108,7 +108,7 @@ func (a *Admin) validate() error {
 		return util.NewValidationError("please set a password")
 	}
 	if !config.SkipNaturalKeysValidation && !usernameRegex.MatchString(a.Username) {
-		return util.NewValidationError(fmt.Sprintf("username %#v is not valid, the following characters are allowed: a-zA-Z0-9-_.~", a.Username))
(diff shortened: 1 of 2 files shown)


drakkan:
Hi,

have you tried to set `skip_natural_keys_validation` to `true` [here](https://github.com/drakkan/sftpgo/blob/main/sftpgo.json#L192)? Please search for `skip_natural_keys_validation` [here](https://github.com/drakkan/sftpgo/blob/main/docs/full-configuration.md) to understand the rationale. Thanks

JohnTheNerd:
I understand, thank you for your help!

PR #554, opened 2021-09-23
Added email field for user account

rashmimaharana:
Hi @drakkan 

I have added email field for user account and have done respective changes in sql. Please verify and let me know if you have any issue, i will do those changes.

Thanks,
Rashmi Maharana

### DIFF
--- a/httpd/httpd_test.go
+++ b/httpd/httpd_test.go
@@ -10772,6 +10772,7 @@ func TestWebUserAddMock(t *testing.T) {
 	user.UID = 1000
 	user.AdditionalInfo = "info"
 	user.Description = "user dsc"
+	user.Email = "test@test.com"
 	mappedDir := filepath.Join(os.TempDir(), "mapped")
 	folderName := filepath.Base(mappedDir)
 	f := vfs.BaseVirtualFolder{
@@ -10788,6 +10789,7 @@ func TestWebUserAddMock(t *testing.T) {
 	form := make(url.Values)
 	form.Set(csrfFormToken, csrfToken)
 	form.Set("username", user.Username)
+	form.Set("emai", user.Email)
 	form.Set("home_dir", user.HomeDir)
 	form.Set("password", user.Password)
 	form.Set("status", strconv.Itoa(user.Status))
@@ -11090,6 +11092,7 @@ func TestWebUserUpdateMock(t *testing.T) {
 	user.AdditionalInfo = "new additional info"
 	form := make(url.Values)
 	form.Set("username", user.Username)
+	form.Set("email", user.Email)
 	form.Set("password", "")
 	form.Set("public_keys", testPubKey)
 	form.Set("home_dir", user.HomeDir)
--- a/dataprovider/dataprovider.go
+++ b/dataprovider/dataprovider.go
@@ -1608,6 +1608,9 @@ func validateBaseParams(user *User) error {
 	if !filepath.IsAbs(user.HomeDir) {
 		return util.NewValidationError(fmt.Sprintf("home_dir must be an absolute path, actual value: %v", user.HomeDir))
 	}
+	if user.Email != "" && !emailRegex.MatchString(user.Email) {
+		return util.NewValidationError(fmt.Sprintf("email %#v is not valid", user.Email))
+	}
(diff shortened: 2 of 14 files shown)


drakkan:
Thanks but you should prepare a database migration, otherwise email will be available only for new installations and database queries fail if someone upgrade an old installation.

Email should also be validated as we do for admins otherwise we accept any string, please take a look [here](https://github.com/drakkan/sftpgo/blob/main/dataprovider/admin.go#L216)

No need to update sftpgo_api_cli, it is deprecated and it only works with v1.2.x and before.

rashmimaharana:
Thanks @drakkan 
Doing those changes and will update soon.

rashmimaharana:
Hi @drakkan 

I have done the changes mentioned. Please review once.

Thanks,
Rashmi Maharana

drakkan:
Hi,

you have to create a new migration, not editing an existing one, otherwise users already at migration 12 will have issues (many users use the development branch to test/use new features). 

I think a user should be allowed to edit its own email, maybe we should rename the "Credentials" page to "Profile" or something similar, however please complete this PR first. Thanks

rashmimaharana:
Thanks @drakkan 
Got you point. Doing those changes.

rashmimaharana:
Hi @drakkan 

Its done. Please review once.

Thanks,
Rashmi Maharana

httpd/httpd_test.go
```diff
@@ -11090,6 +11092,7 @@ func TestWebUserUpdateMock(t *testing.T) {
 	user.AdditionalInfo = "new additional info"
 	form := make(url.Values)
 	form.Set("username", user.Username)
+	form.Set("email", user.Email)
```
drakkan, line 11095: please add a check to verify that the email was really set, and do the same also if the email is set using the REST API

drakkan:
it looks better, thanks. Let's see if CI passes

https://github.com/drakkan/sftpgo/actions/runs/1269832460

drakkan:
There are a lot of warnings like this one:

```
2021-09-24T12:57:05.315 WRN database version 13 is newer than the supported one: 12
```

no time to investigate now, but probably something is wrong

rashmimaharana:
2021-09-24T12:57:05.315 WRN database version 13 is newer than the supported one: 12

This because i have added value 13 in DB upgrade function and its updating schema_version table with value as 13.
But not sure why its falling under "default" block of switch and giving the above warning.

coming to MySQl, i had missed the closing bracket in users table by mistake, so might be that's why it was failling. I have added  that missing closing bracket.

image
image

drakkan:
Try to set database version to 13 [here](https://github.com/drakkan/sftpgo/blob/main/dataprovider/sqlcommon.go#L22)

rashmimaharana:
@drakkan 
Done , updated to 13. Thanks.

drakkan:
Thank you. I think if you add some test cases to check the email field they will fail with the memory provider. 

I can fix this myself. If I have time this week end, I'll try to merge your patch with some minor modifications. Otherwise I'll work on this patch the next weekend.

The email field will be used in future updates for password recovery, to send data retention deletion reports or things like this.

I don't like the table you added to the UI, I'll try to rework that page to make the email field editable (maybe if the admin grants a specific permission).

rashmimaharana:
thanks @drakkan 
i will be happy to add changes. Please let me know, will be happy to do.

drakkan:
it is enough for now, thanks. I will be working on data retention this weekend, and will try to review this PR if there is some time left, possibly I will ask you to change something after a more thorough review, thank you very much appreciated

drakkan:
Hi,

I merged this patch after fixing some minor issues and removing the change for the credential page. I'll try to refator that page the next week. 

Obviously you remain the author of the commit

https://github.com/drakkan/sftpgo/commit/0661876e99f4691084e2175985a602d41ff771ce

thank you for your contribution, very appreciated

rashmimaharana:
Thanks @drakkan for all your help and direction.

drakkan:
thanks to you for the contribution, I'm working on a new profile page, it will look like this

Schermata del 2021-09-28 10-53-15

PR #564, opened 2021-10-01
Flesh out examples/ldapauth

jmkristian:
Support 'virtual' users who have no homeDirectory, uidNumber or gidNumber.
Permit read-only access by a user named "anonymous", with any password.
Assume a conventional DIT with users under ou=people,dc=example,dc=com.
Read the LDAP bindPassword from a file (not baked into the code).
Log progress and problems to syslog.

### DIFF
--- a/examples/ldapauth/main.go
+++ b/examples/ldapauth/main.go
@@ -4,17 +4,22 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"log"
+	"log/syslog"
 	"os"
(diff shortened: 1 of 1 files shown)


drakkan:
Hi,

thank you! In v2.2 I introduced a plugin system, I plan to add a generic authentication plugin that supports ldap and system users (no definite schedule for now). I don't have a real LDAP test system, if you are interested I can ping you once the basic plugin works and you could help me with the LDAP part.

PR #679, opened 2022-01-15
systemd-security: add some easy wins

lpcvoid:
We can tighten security by adding the following to the systemd service file. Please tell me if my assumptions are wrong somewhere - I tested this with my local setup and it works fine, this should not compromise any functionality while removing potentially hazardous capabilities.

* NoNewPrivileges: should never be needed
* DevicePolicy: only basics required
* PrivateDevices: only needs mounted stuff for file access anyhow, never devices or memory or whatever
* ProtectSystem: no need to change boot or other critical files
* RestrictAddressFamilies: INET and UNIX only, no need for any other address families

Thanks for this project, it's amazing.

Cheers!

### DIFF
--- a/init/sftpgo.service
+++ b/init/sftpgo.service
@@ -17,6 +17,11 @@ KillMode=mixed
 PrivateTmp=true
 Restart=always
 RestartSec=10s
+NoNewPrivileges=yes
+PrivateDevices=yes
+DevicePolicy=closed
+ProtectSystem=true
+RestrictAddressFamilies=AF_INET AF_INET6 AF_UNIX
 
 [Install]
 WantedBy=multi-user.target


init/sftpgo.service
```diff
@@ -17,6 +17,11 @@ KillMode=mixed
 PrivateTmp=true
 Restart=always
 RestartSec=10s
+NoNewPrivileges=yes
+PrivateDevices=yes
+DevicePolicy=closed
+ProtectSystem=true
+RestrictAddressFamilies=AF_INET AF_INET6
```
drakkan, line 24: please add `AF_UNIX`, the REST API can be configured to run over UNIX domain sockets

drakkan:
Hi,

thanks for your patch, it looks good, only add AF_UNIX. 

Please also try to run SFTPGo as root and then set a UID/GID != 0 for some users and check that SFTPGo can impersonate the specified UID/GID.

This contribution is really appreciated, thanks

init/sftpgo.service
lpcvoid: Good point, done.

lpcvoid:
Good idea, will test with user impersonation and report back.

lpcvoid:
I ran as root (service) and then used the web UI to set user UIDs and GIDs to existing IDs, and then copied a file over sftp. It chown'ed it correctly to the user/group set in the web UI, with 664 permissions. That seems to work as expected.

drakkan:
Thanks!

drakkan:
Hello,

https://cla-assistant.io/drakkan/sftpgo?pullRequest=679

You haven't contributed any SFTPGo code, you improved the systemd unit, but we still appreciate if you can sign the CLA.

Thank you!

Issue #204, opened 2020-10-25
Run as subsystem


d4z3x:
Hi there,

First off, amazing work! And the help and pointers in another repo!

Now, for the point... I've combed through the docs. But I can't find a way to run this as a subsystem through sshd. Using sshd, I need to force members of a specific LDAP/IPA group in Linux to launch the sftpgo binary (much like how the bundled sftp-server works). 

Is this possible? With a subsystem, no need for an auth check or use SSHD inside of soft-go. And run on default port 22-ssh (corp firewall only allows 22)... 

Side note:
Mapping users is almost perfect, and I need to map each user coming in to a system user. I suppose for now, to keep things simple, I can add the users through the web interface, and map the UID to the system user or via the hook run a chmod on an upload/etc event.

drakkan:
Hi,

running SFTPGo as SSH subsystem is not currently supported. I could investigate how to add such support.

If you can run SFTPGo as system service it should be able to handle just fine your use case: you only have to write an external authentication hook that returns the user with the correct home dir and uid/gid

farhany:
> Hi,
> 
> running SFTPGo as SSH subsystem is not currently supported. I could investigate how to add such support.

Would need to plumb io.Stdout, io.Stdin and disable ssh. Thanks for this. I will take a look as well, and see if I can help.
 
> If you can run SFTPGo as system service it should be able to handle just fine your use case: you only have to write an external authentication hook that returns the user with the correct home dir and uid/gid

The problem is that the we need to use port 22 for regular domain authentication, so that wouldn't be a plugin solution, sadly... at this point.

Hooks are pretty awesome.

drakkan:
> > Hi,
> > running SFTPGo as SSH subsystem is not currently supported. I could investigate how to add such support.
> 
> Would need to plumb io.Stdout, io.Stdin and disable ssh. Thanks for this. I will take a look as well, and see if I can help.

Yes, for what I understand, we basically need to wrap `os.Stdin` and `os.Stdout` inside an `io.ReadWriteCloser` and then pass this wrapper to [NewRequestServer](https://github.com/pkg/sftp/blob/master/request-server.go#L51) and it should work.

We could add a new subcommand to sftpgo, something like this:

```
sftpgo sftp-subsystem -u <username> -h <home dir> and so on
```

I'll try to take a look as soon as I have some free time, probably the next weekend, thank you

> 
[...]

PR #704, opened 2022-01-31
add build flag to entirely disable journald logger

jeremy-clerc:
The library used to log to journald go-systemd always creates
a socket on init, no matter if you'll use it or not:

https://github.com/coreos/go-systemd/blob/9ed442a88da2e6230e1d13be1820a5357055aec8/journal/journal_unix.go#L56

Journald logging is "eventually" available only in the subsystem
and only if the user decides to activate.

No matter which part of sftpgo you use, there is always somewhere
where you include the sftpgo/logger, even `sftpgo help` and so
you always get a socket.

This change adds a build flag so sftpgo can be built without
go-systemd/journal dependency and so prevent the creation of
a socket that will not be used.

### DIFF
--- a/docs/build-from-source.md
+++ b/docs/build-from-source.md
@@ -8,6 +8,7 @@ The following build tags are available:
 - `nos3`, disable S3 Compabible Object Storage backends, default enabled
 - `noazblob`, disable Azure Blob Storage backend, default enabled
 - `nobolt`, disable Bolt data provider, default enabled
+- `nojournald`, disable journald logger, default enabled
 - `nomysql`, disable MySQL data provider, default enabled
 - `nopgsql`, disable PostgreSQL data provider, default enabled
 - `nosqlite`, disable SQLite data provider, default enabled
--- a/logger/journald.go
+++ b/logger/journald.go
@@ -1,5 +1,5 @@
-//go:build linux
-// +build linux
+//go:build linux && !nojournald
+// +build linux,!nojournald
 
 package logger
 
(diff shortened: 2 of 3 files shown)


drakkan:
Thanks for noticing. Before merging this I want to explore if there is some better replacement (maybe something working also on windows) or if the upstream library will accept a PR to add a less hacky initialization method.

Subsystem support was added to close #204 but I never received a real feedback (the user that requested the feature is mainly interested to copy the code and develop its own solution) so I'm not concerned about backward compatibility

drakkan:
[This library](https://github.com/ssgreg/journald) could be a good replacement. I'll try it as soon as I'll have some free time. I also have to write a zerolog wrapper but this is not a big issue

drakkan:
Hi,

can you please test [this patch](https://github.com/coreos/go-systemd/pull/397)

a replace like this should be enough

```
github.com/coreos/go-systemd/v22 => github.com/drakkan/go-systemd/v22 v22.0.0-20220201123313-e7313bc7cda4
```

The zerolog adapter call Enabled before each log (this could be optimized too) and so it does not require the implicit initialization. 

Maybe commenting on the upstream PR could help too. Thanks

jeremy-clerc:
The patch works, thank you.
Closing this PR and I'll use your patch until it is the default.

Thanks!

drakkan:
Hi,

the patch was merged upstream and it is now included in both the SFTPGo development version and in the 2.2.x branch. The version on the 2.2.x branch will be likely tagged as v2.2.2 this weekend. A confirmation that these versions fix the issue would be appreciated, thanks

jeremy-clerc:
I'll have a look asap and let you know

jeremy-clerc:
Looks good, thanks!

drakkan:
Thank you

PR #744, opened 2022-02-25
Deny file rename in directories with Upload/Delete permissions and missing rename permission

derekbking:
I believe there is a bug here in the connection class. It seems that users will be allowed to rename files in the scenario that they have delete/upload permission to a sub directory, but do not have rename permissions.

From my local testing it seems that omitting the rename permission still allows renames to happen. Please let me know what you think.

### DIFF
--- a/common/connection.go
+++ b/common/connection.go
@@ -830,11 +830,11 @@ func (c *BaseConnection) hasRenamePerms(virtualSourcePath, virtualTargetPath str
 		return c.User.HasPerm(dataprovider.PermCreateDirs, path.Dir(virtualTargetPath))
 	}
 	// file or symlink
-	if c.User.HasPerm(dataprovider.PermRenameFiles, path.Dir(virtualSourcePath)) &&
-		c.User.HasPerm(dataprovider.PermRenameFiles, path.Dir(virtualTargetPath)) {
-		return true
-	}
 	if !c.User.HasAnyPerm([]string{dataprovider.PermDeleteFiles, dataprovider.PermDelete}, path.Dir(virtualSourcePath)) {
+		return false
+	}
+	if !c.User.HasPerm(dataprovider.PermRenameFiles, path.Dir(virtualSourcePath)) ||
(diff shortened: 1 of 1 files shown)


drakkan:
Hi,

I'm open to discussions, this behaviour is intentional (as you can see some rename test case fail now). 

With the current implementation, a rename is allowed if:

1) the user has rename permission on both source and target path
2) the user has delete permission on source path and create/upload on target path

I initially added this because renames aren't always atomic, for example renames are a two step operations for cloud based backends and for local providers if the rename happen between different [filesystems](https://github.com/drakkan/sftpgo/blob/main/vfs/osfs.go#L104)

My assumption could be opinable

derekbking:
Hello,

Thank you for the quick response. It makes sense how upload/create and delete permissions are effectively like having rename permissions and how this simplifies the two-step operations.

Our application does not respond well to rename operations under some circumstances. Some clients such as WinSCP initially upload a file as filename.filepart and immediately renames the file after the upload is complete. This causes a scenario where we attempt to do work on a file based on the upload complete event. The problem is that if the client immediately renames the file when we go to do the work, we have no file. We have configuration within our app to support this scenario (they can treat files with .filepart as waiting for a rename effectively). The problem is that some users will not know to do this. In this scenario, we wanted to deny renames, but still allow create/upload.

It feels there is a distinction to be made between the upload/create permissions and delete permissions. Although, I understand how this can complicate implementation and this issues may be specific to my use case.

drakkan:
Well, I think if we only allow renaming if the rename permission is granted on both the source and the target paths, the code will be simplified. Please let me think about this a few days. I'll also read the docs  for similar proprietary software to understand how they handle the rename permission. I have probably introduced an unnecessary complication in the SFTPGo code

derekbking:
Thank you. Your help is much appreciated!

drakkan:
Implemented in https://github.com/drakkan/sftpgo/commit/b64d3c2fbfe9388e5a57319f5fc197884d4bc5b8

PR #761, opened 2022-03-16
add global allow from list

cgroschupp:


cgroschupp:
In the feature list I found [this](https://github.com/drakkan/sftpgo/blob/main/README.md?plain=1#L39):
> Per-user and global IP filters: login can be restricted to specific ranges of IP addresses or to a specific IP address.
I found the per-user part inside the user acl settings, but i don't found any reference to the global ip filter.

I want to use sftpgo in a QA environment and I don't want a user to be able to configure sftpgo to be reachable only as I have previously defined in the global IP filter. The list of IPs is about 10-20 entries long.

drakkan:
> In the feature list I found [this](https://github.com/drakkan/sftpgo/blob/main/README.md?plain=1#L39):
> 
> > Per-user and global IP filters: login can be restricted to specific ranges of IP addresses or to a specific IP address.
> 
> I found the per user part in the user ACL settings, but I didn't find any reference to the global IP filter.
> 
> I want to use sftpgo in a QA environment and I don't want a user to be able to configure sftpgo to be reachable only as I have previously defined in the global IP filter. The list of IPs is about 10-20 entries long.

Hi,

for global IP filters I mean the defender [blocklist](https://github.com/drakkan/sftpgo/blob/main/docs/defender.md). It uses a trie for lookups and so is very fast even with many entries.

Let me the weekend to think about this addition, I have some concerns:

1) a long term goal is to limit the number of settings moving them from the configuration file to the database
2) users could add many entries in this global list and lookups could be slow
3) we can define a global blocklist in the defender section and a global allow list outside the defender, this could be confusing

For your use case will you set `allow_from_additive` to `true` or to `false`?

drakkan:
Ok, I checked some proprietary products similar to SFTPGo and indeed they have a blocklist and safelist for their defender implementation but also a global whitelist that restricts access to only listed IPs/networks, all other client connections will be dropped before they even try to authenticate.

I think we can add something similar (so without the merging feature) and we can do this check globally for all the configured services [here](https://github.com/drakkan/sftpgo/blob/main/common/common.go#L927). I would also prefer a trie based lookup (as we already do for safelist/blocklist) so this list could be large without degrading performance.

If you want to adapt your PR is appreciated, otherwise I'll implement this feature myself, thanks for the PR and the idea.

drakkan:
Please give a try to the latest commit https://github.com/drakkan/sftpgo/commit/d9f30e7ac57f56260ede2ecd192c4ecd708f680d

thanks

cgroschupp:
@drakkan perfect, thank for your help.

PR #774, opened 2022-03-25
Add role field array support

ismailbaskin:
Some OIDC providers like AAD provides `roles` field with array type.

### DIFF
--- a/httpd/oidc.go
+++ b/httpd/oidc.go
@@ -208,7 +208,7 @@ func (t *oidcToken) isAdmin() bool {
 		return v == "admin"
 	case []interface{}:
 		for _, s := range v {
-			if s.(string) == "admin" {
+			if val, ok := s.(string); ok && val == "admin" {
 				return true
 			}
 		}
--- a/httpd/oidc.go
+++ b/httpd/oidc.go
@@ -160,18 +160,18 @@ func newOIDCPendingAuth(audience tokenAudience) oidcPendingAuth {
 }
 
 type oidcToken struct {
-	AccessToken  string   `json:"access_token"`
-	TokenType    string   `json:"token_type,omitempty"`
-	RefreshToken string   `json:"refresh_token,omitempty"`
-	ExpiresAt    int64    `json:"expires_at,omitempty"`
-	SessionID    string   `json:"session_id"`
-	IDToken      string   `json:"id_token"`
-	Nonce        string   `json:"nonce"`
-	Username     string   `json:"username"`
-	Permissions  []string `json:"permissions"`
-	Role         string   `json:"role"`
-	Cookie       string   `json:"cookie"`
-	UsedAt       int64    `json:"used_at"`
+	AccessToken  string      `json:"access_token"`
+	TokenType    string      `json:"token_type,omitempty"`
+	RefreshToken string      `json:"refresh_token,omitempty"`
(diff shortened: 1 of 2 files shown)


httpd/oidc.go
```diff
@@ -203,7 +203,19 @@ func (t *oidcToken) parseClaims(claims map[string]interface{}, usernameField, ro
 }
 
 func (t *oidcToken) isAdmin() bool {
-	return t.Role == "admin"
+	switch v := t.Role.(type) {
+	case string:
+		return v == "admin"
+	case []interface{}:
+		for _, s := range v {
+			if s.(string) == "admin" {
```
drakkan, line 211: Please change to something like this:

`if val, ok := s.(string); ok && val == "admin" {`

otherwise we'll get a panic if the slice contains something other than strings

httpd/oidc_test.go
```diff
@@ -1005,3 +1005,28 @@ func getPreLoginScriptContent(user dataprovider.User, nonJSONResponse bool) []by
...
+		input interface{}
+		want  bool
+	}
+
+	emptySlice := make([]interface{}, 0)
+
+	tests := []test{
+		{input: "admin", want: true},
+		{input: append(emptySlice, "admin"), want: true},
+		{input: append(emptySlice, "user", "admin"), want: true},
```
drakkan, line 1020: add this

`{input: append(emptySlice, 1), want: false},`

it will crash with the current code and it will work after the suggested change

drakkan:
Hi,

I had thought of the role field as a custom (string) field, but it is okay to use the provider built-in roles as well, please change the PR as required and I'll merge it, thanks!

drakkan:
Out of curiosity, do your provider support the end session endpoint? Do you see [this log](https://github.com/drakkan/sftpgo/blob/main/httpd/oidc.go#L128)? Thanks

ismailbaskin:
> Out of curiosity, does your provider support the end session endpoint? Do you see [this log](https://github.com/drakkan/sftpgo/blob/main/httpd/oidc.go#L128)? Thanks

Yes, [it does](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc#send-a-sign-out-request). And, I see.

```
{"level":"debug","time":"2022-03-25T09:30:47.860","sender":"httpd","message":"oidc end session endpoint \"https://login.microsoftonline.com/[REDACTED]/oauth2/v2.0/logout\""}
```

drakkan:
Thanks!

ismailbaskin:
I would like to thank you for bringing such a magnificent project to life and accepting this PR.

drakkan:
The project is growing much more than I initially expected thanks to the external contributions of ideas and code and the kind companies and people who are sponsoring my work. I don't use SFTPGo in production myself but it seems useful to many people and companies.
This is the power of the open source. I guess a similar proprietary project would not have had the same growth.

There are also bad things like people demanding immediate and free support, but I can always avoid answering if they are too annoying :smile:

drakkan:
Hello,

https://cla-assistant.io/drakkan/sftpgo?pullRequest=774

Thank you!

drakkan:
Hello @ismailbaskin 

https://cla-assistant.io/drakkan/sftpgo?pullRequest=774

Thank you!

PR #782, opened 2022-04-01
PD5 Support

ar-comlog:
OLD != BAD. MD5 Support for passwords. It makes switching easier!


drakkan:
At most I can accept read support just for compatibility, but saving password in md5 ourself is out of question. md5 has no security

Issue #762, opened 2022-03-17
[Feature request] Event Manager


lublak:
Currently it is only possible to define the hooks via config.
Thereby the server must be restarted.
It would be practical to set the events via the interface and to define multiple urls (also via the interface).

drakkan:
Hi,

I have a temporal undefinite plan to simply the configuration by moving some settings from the config file to the data provider so we can easily set them from the web ui, but this will likely happen after v2.3, I currently have several other requets from project sponsors.

Can you please better explain what do you mean when you write "and to define multiple urls (also via the interface)"?
The events contains the info about the user so one endpoint is generally enough

lublak:
Hi @drakkan,

thanks for the very fast answer.
Here is roughly what I meant. I think pictures say more than a thousand words.

![grafik](https://user-images.githubusercontent.com/44057030/158819022-402be4ef-8119-4017-bb24-7760d84f1824.png)
![grafik](https://user-images.githubusercontent.com/44057030/158820546-46bff62e-f434-41c0-a204-7d2338ae01ac.png)

Thus, different events can be defined with different hooks.

drakkan:
Hi,

honestly this seems excessive, I feel the current configurability is enough. If you want this level of customization you can write your own eventsreceiver plugin.

Additionally I don't want to create a separate page for each configuration (one for hooks, another one for max connections etc.), I want a single server configuration page divided in different sections and move them some configurations that can be applied without a restart

drakkan:
There is no way to extend the interface within a plugin or to write to existing database. The existing plugins use a separate database.

You could simply notify the events to a single plugin/hook and then multiplex from there. 

I can understand that you want to be able to change a hook url without restarting the server but I cannot understand why you need to forward, for example, upload events to an endpoint and download events to another one. SFTPGo cannot support any particular use case out of there, however, as explained above you could write a simple wrapper that receive and forwards the events (using a plugin or a hook).

Are you trying to expose SFTPGo to external admins? Do you want to integrate SFTPGo with a legacy application that accepts different events on different URLs and you don't want to change this application?

lublak:
@drakkan 
I want to use SFTPGo events with n8n webhooks.
That means I would have to enter a new address in sftpgo for each workflow with a webhook.
But of course I don't want to have another webfrontend for every single mini system, so I want to extend sftpgo then.

But I also understand that you don't want the system to be more complicated than it is.
I think I will abandon the idea and develop a plugin with its own frontend and login.
[...]

PR #818, opened 2022-05-05
docs(cn): support translation for Simplified Chinese

HerbertHe:


drakkan:
Hi,

I appreciate this PR, but I am unable to maintain the translation for Simplified Chinese myself, so it may become out of date when I make changes to the English version

HerbertHe:
I can help for Chinese translation, just need to let me know via at me~

And maybe I would contribute to web interface for i18n support, because some of users don't speak English💪

drakkan:
> I can help for Chinese translation, just need to let me know via at me~
> 

thanks

> And maybe I would contribute to web interface for i18n support, because some of users don't speak Englishmuscle

To support localizations for the webui we need a proper translation system in place. If you just translate the templates a lot of strings will remain in english, it is something I would like to add soon or later

HerbertHe:
@drakkan 
I am a frontend developer and gopher. For i18n support, I suggest using  frontend schema like i18n-next.

drakkan:
> @drakkan I am a frontend developer and gopher. For i18n support, I suggest using frontend schema like i18n-next.

I'm a backend developer, I never used i18n-next, I'll try to study it.

All the validation errors are generated in Go code, for example take a look [here](https://github.com/drakkan/sftpgo/blob/main/dataprovider/dataprovider.go#L1944).

I think we need a system that is able to translate both the text in the Go code and in the templates. I think `golang.org/x/text/language` and `golang.org/x/text/message` should work, but I haven't looked at them in depth yet.

For now the translation is only an idea without a definite schedule, I think it will require a lot of time

HerbertHe:
Right, if the web interface could be refactored as frontend project by using React and RESTFUL API, we can support i18n easier.

But it will spend a lot time to meet the goal😐

HerbertHe:
Or, we can provide a configuration option `lang` and support custom language file formated like `en.json`. Then we can load the language file to map where they should be.

HerbertHe:
But if we only support `lang` option, we can't support multiple languages switching. So, maybe frontend via JavaScript loading is the best solution.

HerbertHe:
Maybe, we can switch languages via params from routes, like `/en/home` and `/zh/home`. Instead, from query, `?lang=zh`.

If the program could not get the specific language, then fallback to default language like English.

But the built-in language packages would cause the project more and more large... But we can meet the goal only by the backend.

drakkan:
I'm not sure yet about the approach. I want to finalize v2.3.0 and then work on #762. After these developments I'll try to find the time to better evaluate the multi-launguage feature

PR #826, opened 2022-05-10
Add flag to indicate if protocol is secure to hooks

rndb8418:
The aim of this is to add the ability to allow/deny access based on if they are connecting with a 'secure' protocol, i.e. if it is FTP or FTPS, or HTTP vs HTTPS.

Use case: if supporting legacy software that only supports FTP we would want to allow the connection for plaintext FTP but deny plaintext FTP for all other users.

At the moment this is not possible (from what I can see), as only the protocol (HTTP, FTP, SFTP...) is passed in the pre-login hook without any detail around the client connection.

There might be a better way of achieving this with more detail passed to the hook, but thought this would suit as a proof of concept initially.

If this seems like a reasonable change, I think it would also be worthwhile adding to the other hooks (connection / external auth) where applicable. Understandably some won't be able to determine if it's secure at time of hook (connection hooks for explicit FTPS would likely show as insecure if the hook occurs before TLS negotiation).

I wasn't able to get the test suite to run fully to check if these changes are working as expected, is there any docs on getting this set up for a dev environment?

### DIFF
--- a/dataprovider/dataprovider.go
+++ b/dataprovider/dataprovider.go
@@ -957,20 +957,20 @@ func CheckCachedUserCredentials(user *CachedUser, password, loginMethod, protoco
 
 // CheckCompositeCredentials checks multiple credentials.
 // WebDAV users can send both a password and a TLS certificate within the same request
-func CheckCompositeCredentials(username, password, ip, loginMethod, protocol string, tlsCert *x509.Certificate) (User, string, error) {
+func CheckCompositeCredentials(username, password, ip, loginMethod, protocol string, protocolSecure bool, tlsCert *x509.Certificate) (User, string, error) {
 	username = config.convertName(username)
 	if loginMethod == LoginMethodPassword {
-		user, err := CheckUserAndPass(username, password, ip, protocol)
+		user, err := CheckUserAndPass(username, password, ip, protocol, protocolSecure)
 		return user, loginMethod, err
 	}
-	user, err := CheckUserBeforeTLSAuth(username, ip, protocol, tlsCert)
+	user, err := CheckUserBeforeTLSAuth(username, ip, protocol, protocolSecure, tlsCert)
 	if err != nil {
 		return user, loginMethod, err
 	}
 	if !user.IsTLSUsernameVerificationEnabled() {
 		// for backward compatibility with 2.0.x we only check the password and change the login method here
 		// in future updates we have to return an error
-		user, err := CheckUserAndPass(username, password, ip, protocol)
+		user, err := CheckUserAndPass(username, password, ip, protocol, protocolSecure)
 		return user, LoginMethodPassword, err
 	}
 	user, err = checkUserAndTLSCertificate(&user, protocol, tlsCert)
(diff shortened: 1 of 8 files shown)


drakkan:
Hi,

this feature has been requested in the past by a company that has considered sponsoring the project (unfortunately the AGPL license was not acceptable for them), so I have already done some analysis in the past.

I think this check might be useful for FTP, I'm not sure for HTTP. Either way your HTTPS check is wrong, it won't work behind a proxy. 

Your patch only checks for the FTP control connection, we should also check for the transfer connection and the check should not be delegated to an external hook but we should add this configurability directly in ftpserverlib and add an option to SFTPGo. 
From the admin UI or from the hook we can then enable this restriction on a per-user basis.

rndb8418:
Hi,

Thanks for taking a look so quickly. For HTTP, that makes sense as it's best to keep users on secure connections where possible, and I can't think of a good use case where a user would _need_ to use plain HTTP.

For FTP, the data connection won't be established unless it's needed for passing of data due to specific commands on control connection and this wouldn't be available at login. Ideally a user would be rejected at login if they should _not_ use plaintext FTP if that's set on their account.

If **userA** was allowed plaintext FTP, but **userB** must have secure connection to login, I would think:

**userA via plaintext FTP**
1. Connects to SFTPGo instance via FTP
2. Attempts to login without encryption (i.e. `USER test`)
3. Login allowed (if correct credentials) on plaintext FTP

**userB via plaintext FTP**
1. Connects to SFTPGo instance via FTP
2. Attempts to login without encryption (i.e. `USER test`)
3. Login rejected as they must use secure connection

**userA OR userB via FTPS**
1. Connects to SFTPGo instance via FTP
2. Issues `AUTH TLS` command to encrypt control connection
3. Attempts to login (i.e. `USER test`)
5. Login allowed (if correct credentials) on FTPS

I think there would be a use-case for enforcing data connection encryption too, but that isn't possible at point of login. Having a 'require secure connection' flag on the user account could make it so plaintext logins & plaintext data connections are rejected?

I think these would be rejected once the username is known, prior to password command, if that user does not allow insecure connections. That would prevent the password being transmitted in plaintext. This makes it unsuitable for a pre/post-login hook as by then the full credentials are already transmitted I believe.

drakkan:
when the pre-login hook is executed we have already received username **and** password, ftpserverlib calls the AuthUser callback after the `PASS` command.

A while ago I added a setting to require TLS to [ftpserverlib](https://github.com/fclairamb/ftpserverlib/pull/169). This setting enforces the TLS requirement when the `USER` command is sent (`PASS` command is sent after the USER command) and whenever a transfer connection is required (file list, upload, download, ecc). 

This setting is global for the whole the server, I think we should extend this setting so it can be set per-user, so we need to add a new callback to ftpserverlib when the user command is received or something similar. This way we ensure that the password is not trasmitted as plain text if we don't want it. I think this check should be done in ftpserverlib so we can easily enforce the TLS requirement also for the transfer connections, in ftpserverlib there is just a single `TransferOpen` method, in SFTPGo we should check before each command that opens or can open a transfer connection.

This PR cannot be accepted as is, doing this check in the pre-login hook is not ok for the above reasons.

This is a very specific requirement, I'm too busy to work on this or to review a PR for sftpgo/ftpserverlib without a sponsorship sorry

rndb8418:
Thanks I appreciate your time, and I realize the approach taken was incorrect for the problem at hand now. I will look at the `ftpserverlib` to see if I can provide help there for this.

PR #833, opened 2022-05-12
chore: disallow all crawlers with robots.txt

js-timbirkett:
This adds a `/robots.txt` response to disallow all well-behaved bots. This was raised by one of our InfoSec people; I spent a little time disagreeing with them on the security aspect of `robots.txt` as bad actors and bad bots do not care about `/robots.txt`. They say you shouldn't wrestle a pig in mud, you both get dirty, and the pig likes it.


js-timbirkett:
Output is as per Google's spec for `robots.txt`.

Screenshot 2022-05-12 at 10 07 37 PM

drakkan:
thanks!

PR #834, opened 2022-05-13
chore: fix a linting error

js-timbirkett:
I noticed that the `golangci-lint` step was failing on `main`, ran it locally, and have fixed a few things it was raising:
```
❯ golangci-lint run
httpd/httpd_test.go:121:2: `robotsTxtPath` is unused (deadcode)
	robotsTxtPath                   = "/robots.txt"
	^
vfs/statvfs_unix.go:22:18: unnecessary conversion (unconvert)
		Bavail:  uint64(stat.Bavail),
		               ^
vfs/statvfs_unix.go:24:18: unnecessary conversion (unconvert)
		Ffree:   uint64(stat.Ffree),
		               ^
vfs/statvfs_unix.go:25:18: unnecessary conversion (unconvert)
		Favail:  uint64(stat.Ffree), // not sure how to calculate Favail
		               ^
```

Also made use of the existing `healtzPath` value in `TestHealthCheck` for consistency with other tests using the value.

### DIFF
--- a/httpd/httpd_test.go
+++ b/httpd/httpd_test.go
@@ -8905,14 +8905,14 @@ func TestMethodNotAllowedMock(t *testing.T) {
 }
 
 func TestHealthCheck(t *testing.T) {
-	req, _ := http.NewRequest(http.MethodGet, "/healthz", nil)
+	req, _ := http.NewRequest(http.MethodGet, healthzPath, nil)
 	rr := executeRequest(req)
 	checkResponseCode(t, http.StatusOK, rr)
 	assert.Equal(t, "ok", rr.Body.String())
 }
 
 func TestRobotsTxtCheck(t *testing.T) {
-	req, _ := http.NewRequest(http.MethodGet, "/robots.txt", nil)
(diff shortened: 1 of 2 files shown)


drakkan:
Hi,

the first lint fix is ok, the ones in `statvfs_unix.go` are not ok. You are likely on darwin and so Bavail and Ffree are [uint64](https://github.com/golang/sys/blob/master/unix/ztypes_darwin_amd64.go#L92), but on FreeBSD  they are int64 see [here](https://github.com/golang/sys/blob/master/unix/ztypes_freebsd_386.go#L122) and [here](https://github.com/golang/sys/blob/master/unix/ztypes_freebsd_amd64.go#L117).

Please update your patch, thanks

js-timbirkett:
That's interesting. I hadn't realised that the values were different types across platforms.

_Note to self: Sign commits._

drakkan:
> That's interesting. I hadn't realised that the values were different types across platforms.
> 

I discovered this behavior the first time I tried to compile SFTPGo on FreeBSD, I'm not sure why `sys / unix` uses platform dependent types. Thanks for your lint fixes

> _Note to self: Sign commits._

PR #887, opened 2022-06-24
OIDC: Allow Scope to be configured

BobSilent:
In order to request additional information for user creation the OAuth scopes should be configurable and not hardcoded.

Otherwise the existing (default) scopes needs to be extended which would provide the information to other OAuth apps requesting the scope as well. This is not intended.

Please let me know if I missed something.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1370,6 +1370,12 @@ func getHTTPDOIDCFromEnv(idx int) (httpd.OIDC, bool) {
 		isSet = true
 	}
 
+	scopes, ok := lookupStringListFromEnv(fmt.Sprintf("SFTPGO_HTTPD__BINDINGS__%v__OIDC__SCOPES", idx))
+	if ok {
+		result.Scopes = scopes
+		isSet = true
+	}
+
 	roleField, ok := os.LookupEnv(fmt.Sprintf("SFTPGO_HTTPD__BINDINGS__%v__OIDC__ROLE_FIELD", idx))
 	if ok {
 		result.RoleField = roleField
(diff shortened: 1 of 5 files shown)


drakkan:
Hi,

thanks for your contribution. There are some issues:

- scopes are empty if the `scopes` fields are not added to the config file, so users who upgrade from previous versions without upgrading the config file will have problems
- if you set an OIDC parameter from an env var, but not scopes then the scopes will be empty
- openid scope is required, we must error out with a clear error if a user does not set it
- no test cases

Since this is your first contribution, I'll merge this PR as is and I'll fix the issues myself.

BobSilent:
@drakkan Big THANKS for doing the changes directly! :)
For me it was also my first GO contribution and it would be okay if you would have mentioned the missing things. 

A general question: This topic here was for quite clear for me it should be added. But I also have maybe one or two additional topics, how would you like to start them? 
Shall I start with a discussion or an issue? or with a draft PR?

BobSilent:
thanks, I will follow your advice.

Currently I want to evaluate and use it private, but i also have SSO/OAuth solution at home.
Therefore I want to support with my work and PR in a first step.

drakkan:
Hello,

thank you again for this contribution. We have recently updated our contribution policy and we now require signing a CLA. Agreeing to the CLA explicitly states that you are entitled to provide a contribution and that you cannot withdraw permission to use your contribution at a later date. This removes any ambiguities or uncertainties.

If you are unable to sign the CLA, your contribution will be removed.

https://cla-assistant.io/drakkan/sftpgo?pullRequest=887

Thank you!

drakkan:
Thank you for signing the CLA!

PR #960, opened 2022-08-23
Add a new config option for default share expirations

yayayayaka:
This PR adds a new configuration option `share_expiration_days` to
control the date when a newly created share shall expire.

For example, if an administrator declares `share_expiration_days: 30` in
their SFTPGo configuraiton, the expiration date in the "Add a new share"
form would be set to 30 days in the future, as seen in the screenshots below:

2022-08-23T11:18:29+02:00-yunix
2022-08-23T11:18:54+02:00-yunix

The user would still able to choose a different expiration date (or no expiration at all) in the
form:

2022-08-23T11:19:47+02:00-yunix
2022-08-23T11:20:01+02:00-yunix

I am not a Go programmer myself, so this PR might need some polishing. But I hope you get my idea.


drakkan:
Hi,

thanks for your PR.

I have a still undefined plan to reduce the config parameters.

In this case, I think we can add a per-user setting instead of a global config. This way you can, for example, define the preferred expiration on a group and add your users to this group.

yayayayaka:
That would also be fine for me.

yayayayaka:
Thanks!

PR #999, opened 2022-09-28
SFTPGo webdavd 

jatin0123:
How to use sftpgo webdav for multiple domains with their certificates.


drakkan:
I guess you wanted to ask a question and not open a PR.
You can define multiple webdav bindings and each binding can have its own certificate. 

If you, instead, want a name-based virtual hosting simply proxy SFTPGo behind nginx or apache

jatin0123:
could you please share some examples, so I can go through it?

drakkan:
> could you please share some examples, so I can go through it?

No time/motivation sorry, please read the [support policy](https://github.com/drakkan/sftpgo#support-policy)

jatin0123:
I tried multiple binding by adding in it certifcates, it seems that only last binding is able to fetch the certificate and rest of was not getting fetch with cerfticate but i was able to get connected with the webdav.
The error was "Certficate was not issued for this server".
With last binding am not getting any error. If i am doing something wrong, can you please provide some correct example.

Issue #951, opened 2022-08-13
drag and drop upload UI


tooptoop4:
I know I can click the upload button and then navigate on the file upload dialog but would be quicker to drag a file from a folder on desktop and drop somewhere on the browser

PR #1059, opened 2022-11-16
Added simple drag and drop functionality #951

wooneusean:
Addresses shortcoming in #951


wooneusean:
Sorry for the commit spam, never used signoffs before 🙏

drakkan:
Thanks for this patch, web development isn't my favorite skill, so it is much appreciated. 
I had considered things like filepond or dropzone in the past, but I also like this approach and maybe it even provides a better user experience without requiring a javascript library.

Some notes:

- drag a file, click cancel, drag another file, the previous one will remain selected
- if you drag file1 and then file2, file2 will replace file1. 

Another option is to start uploading as soon as a file is dragged into the file list similar to Nextcloud. What do you think about this approach? Nextcloud highlights the files window when dragging and also allows you to upload to a folder if you drag to a folder, but this could be complex

Thanks

wooneusean:
Will address those bugs.

As for uploading as soon as its dropped, that is possible. This will also address the first problem.

Dragging and dropping to a folder directly is complex as you said but I will try something out and see how it goes.

Thanks for the feedback.

wooneusean:
Some considerations have come up when trying to implement drag and drop file uploads and dragging to a specific folder:

1. The original flow where the upload modal will show when dragging over the window cannot be kept because the listener is needed to decide which folder to drop the files at.
2. I have decided to style the file upload input as a dropzone, where a user can drop files or select them by clicking on the dropzone (the usual way). However because of this, it would not be ideal to immediately upload the files as soon as they are selected assuming the same flow is used both the drag and drop and file selection.

I haven't yet figured out the folder drag and drop (mentioned in no. 1) but for point no. 2, I will simply leave it as it has been (no immediate upload) and allow for files to be dragged and dropped. The user needs to click the `Submit` button to upload the files.

Also, you mentioned that dropping another file will replace the old one in your comment. I did some thinking and came to a conclusion that that is the best way in terms simplicity and user experience because, if subsequent files can be added after the first one, there will be a need for a "Remove" or "Clear" button to remove those files. This is completely doable, however I feel that it might be a little bit extra. But if you wanted I could add that.

drakkan:
As stated above, my original plan was to use filepond or dropzone and provide drag and drop there. I have  a working proof of concept using filepond locally. Take a look here

Schermata del 2022-11-17 08-12-57

there is a remove button for each file. 
I also wanted to try dropzone before making a decision, but then there were other priorities and I didn't have time to do it.

If dragging a file into the file browser window opens a popup like this with a single file pre-filled I think that's enough. Users can add or remove files from there.

Forgot about the upload to folder feature, we can add it in the future if required. I don't think we'll ever had something like Nextcloud :smile: . This can only happen if I have enough sponsorship to pay a frontend developer and we're a long way off right now

drakkan:
Just for sharing, here is the diff to have the filepond window you can see above

```
diff --git a/templates/webclient/files.html b/templates/webclient/files.html
index 38c59cc..a1a67f6 100644
--- a/templates/webclient/files.html
+++ b/templates/webclient/files.html
@@ -25,6 +25,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 <link href="{{.StaticURL}}/vendor/datatables/dataTables.checkboxes.css" rel="stylesheet">
 <link href="{{.StaticURL}}/vendor/lightbox2/css/lightbox.min.css" rel="stylesheet">
 <link href="{{.StaticURL}}/vendor/video-js/video-js.min.css" rel="stylesheet" />
+<link href="{{.StaticURL}}/vendor/filepond/filepond.min.css" rel="stylesheet" />
 <style>
     div.dataTables_wrapper span.selected-info,
     div.dataTables_wrapper span.selected-item {
@@ -110,7 +111,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
             </div>
             <form id="upload_files_form" action="{{.FilesURL}}?path={{.CurrentDir}}" method="POST" enctype="multipart/form-data">
                  <div class="modal-body">
-                    <input type="file" class="form-control-file" id="files_name" name="filenames" required multiple>
+                    <input type="file" id="files_name" name="filenames" required multiple>
                 </div>
                 <div class="modal-footer">
                     <input type="hidden" name="_form_token" value="{{.CSRFToken}}">
@@ -229,6 +230,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 <script src="{{.StaticURL}}/vendor/codemirror/codemirror.js"></script>
 <script src="{{.StaticURL}}/vendor/codemirror/meta.js"></script>
 <script src="{{.StaticURL}}/vendor/video-js/video.min.js"></script>
+<script src="{{.StaticURL}}/vendor/filepond/filepond.min.js"></script>
 {{if .HasIntegrations}}
 <script type="text/javascript">
     var childReference = null;
@@ -664,6 +666,12 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
             }
         });
 
+        FilePond.create(document.getElementById("files_name"),{
+            allowMultiple: true,
+            name: 'filenames',
+            required: true
+        });
+
         $('#spinnerModal').on('shown.bs.modal', function () {
             if (spinnerDone){
                 $('#spinnerModal').modal('hide');
@@ -710,7 +718,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
             keepAlive();
             var keepAliveTimer = setInterval(keepAlive, 300000);
 
-            var files = $("#files_name")[0].files;
+            var files = FilePond.find(document.getElementById("files_name")).getFiles();
             var has_errors = false;
             var index = 0;
             var success = 0;
@@ -736,7 +744,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
                     var errorMessage = "Error uploading files";
                     let response;
                     try {
-                        var f = files[index];
+                        var f = files[index].file;
                         var uploadPath = '{{.FileURL}}?path={{.CurrentDir}}'+encodeURIComponent("/"+f.name);
                         var lastModified;
                         try {
@@ -885,7 +893,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
             name: 'addFiles',
             titleAttr: "Upload files",
             action: function (e, dt, node, config) {
-                document.getElementById("files_name").value = null;
+                FilePond.find(document.getElementById("files_name")).removeFiles();
                 $('#uploadFilesModal').modal('show');
             },
             enabled: true
```

This diff is against an old version (2.3.5) but should apply or be easily adapted to the current one. I used filepond 4.30.4

wooneusean:
Roger that, for now will just keep the drag and drop with the prefilled input.

drakkan:
> Roger that, for now will just keep the drag and drop with the prefilled input.
> 
> As for the folder drop uploading, I have come up with an idea but not sure how well it will work but maybe you can try and implement this:
> 
> 1. Select closest table row when hovering over table
> 2. Check if underlying row is a folder
> 3. Get the link of the folder (probably from the anchor tag)
> 4. POST request with files
> 
> A file input is not needed as the files can be retrieved from the `drop` event object from the `dataTransfer.files` attribute.

This is doable in some way, datatables provides a lot of callbacks but let's leave it alone for now, we don't have to copy Nextcloud.

drakkan:
> Roger that, for now will just keep the drag and drop with the prefilled input.
> 

do you plan to add dropzone (v6)? I'm curious to see the differences with filepond

> As for the folder drop uploading, I have come up with an idea but not sure how well it will work but maybe you can try and implement this:
> 
> 1. Select closest table row when hovering over table
> 2. Check if underlying row is a folder
> 3. Get the link of the folder (probably from the anchor tag)
> 4. POST request with files
> 
> A file input is not needed as the files can be retrieved from the `drop` event object from the `dataTransfer.files` attribute.

wooneusean:
> you plan to add dropzone (v6)? I'm curious to see the differences with filepond

I would try to but I am not sure how to, currently I am making changes in my browser dev tools and seeing if it works. Adding dropzone seems like a large undertaking.

drakkan:
I rebased my filepond wip and pushed the changes in this [branch](https://github.com/drakkan/sftpgo/tree/dragdrop). No more time sorry, I have to do my daily work

wooneusean:
I am trying out Dropzone. The way it is written makes it difficult to implement how you implemented the upload function, because each file you saving has its respective POST endpoint. However, Dropzone only allows a single endpoint. I can make it so its purely a holder of files, then use your `$("#upload_files_form").submit()` to trigger the upload, whats your opinion?

drakkan:
> I am trying out Dropzone. The way it is written makes it difficult to implement how you implemented the upload function, because each file you saving has its respective POST endpoint. However, Dropzone only allows a single endpoint. I can make it so its purely a holder of files, then use your `$("#upload_files_form").submit()` to trigger the upload, whats your opinion?

This is exactly what I did in my filepond [WIP](https://github.com/drakkan/sftpgo/commit/945415b6f007cbd48cbbf000bc6703efc29daa64#diff-999942de44fe5fd770b8ced00f400ed68565e4602b7f933799173901040a5292R723)

wooneusean:
Well, I shall leave the decision to you 👍

wooneusean:
I can try to tinker around in the back end to implement multi file uploads, but can I know which file and function handles file upload requests?

drakkan:
> I can try to tinker around in the back end to implement multi file uploads, but can I know which file and function handles file upload requests?

I used multipart uploads in the past. I don't remember the exact issue I had, I switched to upload files one by one as now (you select multiple files and they are uploaded one after the other)

Take a look at the latest commit in the dragdrop branch. I think only minor adjustments are needed

drakkan:
Changes in the dragdrop branch work better than this patch, but there are still issues if you drop a directory. 

If someone wants this feature they have to provide a fully functional patch (starting from the dragdrop branch or from scratch I have no preferences).  Reviewing and iterating or pushing back on PRs is work I can no longer afford to spend all this time on SFTPGo for free

Issue #1088, opened 2022-12-05
Unable to remove per-directory permissions on group via REST API


g-t-u:
When I update secondary group's ACL with REST API call `PUT /groups/{name}`, new per-directory permissions are merged into existing ones: I'm able to add or change them but not delete.

This happens because request's JSON is unmarshalled directly into pre-existing group in `httpd.updateGroup` func.

I submit new PR with same code taken from `httpd.updateUser`.

PR #1089, opened 2022-12-05
Fix updating per-directory permissions in groups.

g-t-u:
A simple fix for #1088


drakkan:
Hi,

I can reproduce the reported issue but I don't think this patch will fix it. Have you tested it? I think other group fields also have the same issue, for example `file_patterns`, and other list based fields. Anyway if you want to contribute to the project please fix the DCO check. Thanks

PR #1112, opened 2022-12-23
Added graceful shutdown description to docker

JonBendtsen:
Describing how to use the graceful shutdown period in a docker SFTPGO container and giving some examples of what happens with both existing and new connections.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -96,9 +96,9 @@ docker logs some-sftpgo
 ### Container graceful shutdown
 
 ```shell
-docker run --name sftpgo.example.net --host sftpgo.example.net \
+docker run --name some-sftpgo \
     -p 2022:2022 \
-	-e SFTPGO_GRACE_TIME=32 \
+    -e SFTPGO_GRACE_TIME=32 \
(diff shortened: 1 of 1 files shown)


drakkan:
Hello,

`--host` seems an unknow docker flag, do you mean `--hostname`?

For uniformity with other examples just use

```
docker run --name some-sftpgo \
    -p 2022:2022 \
    -e SFTPGO_GRACE_TIME=32 \
    -d "drakkan/sftpgo:tag"
```

and please also fix the formatting

Schermata del 2022-12-23 11-32-19

`-e SFTPGO_GRACE_TIME=32` is not properly aligned. Thanks!

drakkan:
Thanks, I'm not sure about the examples:

- the first one shows a failed put. I suppose this means that the client was disconnected after the grace period
- the second one shows an ok upload and then a datetime and a failed connection. I guess this is the output of a script that print the current datetime, connect and upload a file. There is no explicit connection command (`sftp -P 2022 ....`) so this could create confusion.

Maybe we could remove the examples and just describe what happens. I'll think about this later

PR #1113, opened 2022-12-23
Redid the graceful examples

JonBendtsen:
Redid the graceful examples with more details and the command to shutdown the container. Notice that the container only seems to do a 10 second graceful, not 32 seconds?


JonBendtsen:
not done well enough when merging yours and my changes

PR #1114, opened 2022-12-23
Added warning of how to get SFTPGO graceful

JonBendtsen:
Added warning because just using docker stop some-sftpgo will not let SFTPGO decide the graceful shutdown. To do that one has to kill the 1. process (the sftpgo process) inside the container.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -108,6 +108,8 @@ While the SFTPGo container is in graceful shutdown mode waiting for the last con
 
 If no connections are active or `SFTPGO_GRACE_TIME=0` (default value if unset) the container will shutdown immediately.
 
+**Warning**: simply using `docker stop some-sftpgo` will not give you the full graceful shutdown period. To use SFTPGO's graceful period, then run this command: `docker exec some-sftpgo kill 1` which will trigger the graceful shutdown of SFTPGO. If you need to wait for the container to stop, then run `docker exec some-sftpgo kill 1 ; docker wait some-sftpgo`
+
 ### Where to Store Data
 
 Important note: There are several ways to store data used by applications that run in Docker containers. We encourage users of the SFTPGo images to familiarize themselves with the options available, including:


JonBendtsen:
I do wonder if using killall sftpgo is better than using kill 1 because what if you one day create a container where the sftpgo process does not have PID=1 ?

docker/README.md
```diff
@@ -108,6 +108,8 @@ While the SFTPGo container is in graceful shutdown mode waiting for the last con
 
 If no connections are active or `SFTPGO_GRACE_TIME=0` (default value if unset) the container will shutdown immediately.
 
+**Warning**: simply using `docker stop some-sftpgo` will not give you the full graceful shutdown period. To use SFTPGO's graceful period, then run this command: `docker exec some-sftpgo kill 1` which will trigger the graceful shutdown of SFTPGO. If you need to wait for the container to stop, then run `docker exec some-sftpgo kill 1 ; docker wait some-sftpgo`
```
drakkan, line 111: please use `docker kill` to send a signal to a running container instead of executing the `kill` command within the container (not tested but the above command will likely fail with distroless images). 

Some other notes:

- replace **Warning** with `:warning:`
- in sentences like "To use SFTPGO's" please replace SFTPGO with SFTPGo

maybe you start the warning with something like this (English is not my native launguage)

A graceful shutdown is triggered by an interrupt signal so ....

Thanks!

JonBendtsen: Sure. In January after xmas

drakkan:
I did a quick test and `docker stop some-sftpgo` sends SIGTERM and so triggers a graceful shutdown.

If you want to send a reload request (unrelated to graceful shutdown) you have to use something like this

```
docker kill --signal=SIGHUP some-sftpgo
```

or 

```
docker kill --signal=1 some-sftpgo
```

Please reopen providing more info

JonBendtsen:
I actually used podman and not docker, but docker is standard so I used docker commands. Podman should be a drop in replacement.

What I experience is that if I use podman stop, then it kills the connection after 10 seconds and not the graceful period I set.

PR #1115, opened 2022-12-23
Pretty print adding 2. folding layer under Configuration file

JonBendtsen:
Wanted to have a 2. folding layer on the Configuration file to make it easier to read, so this edit is purely pretty print.

Had to add a ## heading 2 for Environment variables at the end of Configuration file segment to clearly indicate that the Configuration file's 2.fold are done.

### DIFF
--- a/docs/full-configuration.md
+++ b/docs/full-configuration.md
@@ -50,11 +50,15 @@ If you don't configure any private host key, the daemon will use `id_rsa`, `id_e
 
 The `gen` command allows to generate completion scripts for your shell and man pages.
 
+## Configuration file
+
 </details>
 
 <details><summary><font size=5> Configuration file</font></summary>
 
 The configuration file contains the following sections:
+
+<details><summary><font size=4>common</font></summary>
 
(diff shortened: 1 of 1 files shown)


docs/full-configuration.md
```diff
@@ -50,12 +50,16 @@ If you don't configure any private host key, the daemon will use `id_rsa`, `id_e
 
 The `gen` command allows to generate completion scripts for your shell and man pages.
 
+## Configuration file
```
drakkan, line 53: why we need this and the other similar headings?

drakkan:
I just pushed https://github.com/drakkan/sftpgo/commit/fe9904a54d5f0fe1f2710f6e0be46acecbf03f19, rebase on top of it if I have missed something. Thanks

docs/full-configuration.md
JonBendtsen: I thought it was prettier with it than without. Especially at the end of the configuration segment to indicate that the sub segment has ended and that we are now once again in the upper segment.

PR #1128, opened 2023-01-03
metrics only available in telemetry server

JonBendtsen:
I do not know which version removed /metrics from the HTTP server, but it does not seem to be available in 2.4.2, so I updated the metrics documentation to reflect this. Replaced with links to telemetry configuration.


drakkan:
Thank you, these doc checks/improvements are really appreciated

drakkan:
Ops test certificates have expired and so test cases fail. I'll fix later today

PR #1129, opened 2023-01-03
Added segment of how to use container health check

JonBendtsen:
Described a few methods of how to use the container health checks with official SFTPGo images.


drakkan:
Hi,

Thanks for the PR.
I think `curl` is installed as dependency in some variant. For other variants such as the distroless one `ps` and `openssl` are not available. 
Also note that the `/healthz` endpoint is available even if telemetry is disabled.
I'll not merge this PR for now. I think I can add a subcommand like `sftpgo ping` so it will work in every docker variant, but I want to think about it a little better first

PR #1131, opened 2023-01-05
How to integrate SFTPGo and ClamAV

JonBendtsen:
This document is an example of how to integrate SFTPGo and ClamAV in a single pod and use clamdscan for every uploaded file.


JonBendtsen:
closing because 1 commit was unsigned

PR #1133, opened 2023-01-06
How to integrate SFTPGo and ClamAV

JonBendtsen:
This document is an example of how to integrate SFTPGo and ClamAV in a single pod such that the Event Manager in SFTPGo could possibly run clamdscan for every uploaded file.

### DIFF
--- a/docs/howto/virus-scanning.md
+++ b/docs/howto/virus-scanning.md
@@ -1,0 +1,206 @@
+# Virus Scanning
+
+It is always good to detect viruses and with the
+[Event Manager](https://github.com/drakkan/sftpgo/blob/main/docs/howto/eventmanager.md) in SFTPGo
+it is easy to scan all uploaded files on the fly immediately once they are uploaded.
+
+## Prerequisite
+
+Before following this example you probably need to be able to run both ClamAV and SFTPGo in a container.
+Please look at the respective documentation for both
+[ClamAV](https://github.com/Cisco-Talos/clamav-documentation/blob/main/src/manual/Installing/Docker.md)
+and [SFTPGo](https://github.com/drakkan/sftpgo/blob/main/docker/README.md).
+It is also a good idea to look at the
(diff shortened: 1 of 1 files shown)


drakkan:
Hello,

thanks for this detailed howto, I haven't yet tested it but I think it's definitely useful but also very specific and will only work with local filesystem. I would prefer to have howto like this as links to external sites (see for example the Active Directory integration or the Traefik howto). Do you have a blog/site where you can publish the howto? Please leave me some more time to think about it better, thanks

JonBendtsen:
I do not have a blog, and I do not intend to start one. What if you made a new folder called user_contributions in the howto folder and then it and similar documents could be put there?

JonBendtsen:
as for only working with local filesystem, I mention that that you can use `--stream` in the :How ClamAV container accesses files in SFTPGo container: segment so ClamAV and SFTPGo does not have to share the filesystem.

Or did you mean Storage backends? True, my description is made with Local storage backend in mind. But what if you in the event manager select `Filesystem events` and also select `Execute sync` wouldn't that be capable of scan before the upload is complete? What happens if the script that scans rejects? I would expect that no part of the file is transferred to SFTPGO are also transferred to the other storage backend, but I suppose that it is a question of write-back or write-through, and it seems like I can not specify either when looking in the docs folder for files for non local storage. (maybe there could be a storage_backend.md file? but that would not contain much, only link to the specific storage backend md files.

drakkan:
My main concern is that people start asking for help/clarifications (for example, see your questions, I think encrypted local filesystem won't work either).
I learned the hard way that once I accept a PR, I have to maintain the feature (or the document in this case) forever, even if the original contributor disappears. 
I appreciate your efforts but I don't have motivation to maintain a such doc sorry, this will take time and work, please don't ask me to invest more for free.

JonBendtsen:
I understand your concerns. What if I posted it under Ideas in the discussion part?

drakkan:
> I understand your concerns. What if I posted it under Ideas in the discussion part?

Sounds good, thanks

PR #1209, opened 2023-02-25
Fix potential ldap injection

tribut:
This fixes an ldap injection in the ldapauth examples.

I can't immediately come up an example how (if at all) this could be exploited.

However, that sentence is a prime candidate for a list of "famous last words", so it should probably be fixed nonetheless.


drakkan:
Thanks!

PR #1230, opened 2023-03-13
Check os permissions as if acting as user

peterverraedt:
This PR proposes to optimize the behaviour of acting as a specified uid/gid for a user, if sftpgo is run as root. We check standard linux permissions as well as extended ACLs. For this we wrap all `pkg.go.dev/os` calls by a package that checks the permissions first before calling the corresponding function. If not on linux, or sftpgo is not running as root, there is a fallback to the default `pkg.go.dev/os` calls.

The main benifits are that file permissions are checked, but syscalls to change uid/gid are avoided.

There are two exceptions of operations that are not being checked that a certain uid/gid has access: The ScanQuota function for virtual folders if triggered by an event, because virtual folders are not one-to-one mapped to a user, but that operation only reads files so it should be fine; and the creation of the user "home folder". The latter could be changed by creating a world-writable directory with the sticky bit set.

In case you would consider to merge this, implementing secondary groups in the database model would be the next step, for situations where folders are shared on gid-basis and users can belong to different groups.

(This is a follow up to #1225)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,6 +44,7 @@ require (
 	github.com/mhale/smtpd v0.8.0
 	github.com/minio/sio v0.3.1
 	github.com/otiai10/copy v1.9.0
+	github.com/peterverraedt/useros v0.1.5
 	github.com/pires/go-proxyproto v0.6.2
 	github.com/pkg/sftp v1.13.6-0.20230213180117-971c283182b6
 	github.com/pquerna/otp v1.4.0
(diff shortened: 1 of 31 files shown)


drakkan:
Thanks for contribution. This seems a potential big changes, I'll try to find the time to test and review it after v2.5 release. Do we really need to support linux ACL? I haven't looked into the code, so this might be a trivial question, but CGO dependency introduction is something to consider carefully. Currently the only CGO dependency is sqlite

peterverraedt:
Sorry for the C-dependency, it is only used in `https://github.com/peterverraedt/useros/blob/main/log.go` for debugging purposes and can easily be removed.

I'll further test and see whether something pops up.

drakkan:
> Sorry for the C-dependency, it is only used in `https://github.com/peterverraedt/useros/blob/main/log.go` for debugging purposes and can easily be removed.
> 

great!

As for logging I would like an interface so we can integrate the library logging with SFTPGo, see the following examples

https://github.com/drakkan/sftpgo/blob/main/internal/logger/lego.go
https://github.com/drakkan/sftpgo/blob/main/internal/logger/hclog.go

ignore this comment if the library logger already allows it

> I'll further test and see whether something pops up.

Thanks!

PR #1282, opened 2023-05-05
Aligned help example with v2.5.0 output

dlehammer:
Suggestion based on the below changelog excerpt

> - Added a CLI command to reset the admin password.
> - Added a CLI command to check if the service is alive.


drakkan:
Thanks

PR #1284, opened 2023-05-08
WebDAV when user authentication fails.

bingo235:
When the password of a user who uses external validation is changed, if the user attempts to validate using the new password, it may fail due to WebDAV caching the user information. In this case, should clear the cache and perform external validation again.


drakkan:
Hello,

I understand the problem. I haven't tested your patch (no time now) but I think it will remove the user from the cache whenever a user submits an incorrect password (even if it hasn't changed) and this will break WebDAV.

bingo235:
When a user submits an incorrect password, the current user cache will be cleared and external validation will be performed again.
There may indeed be a problem of cache penetration.

drakkan:
I think this was already fixed in a different way, Don't remember exactly the commit

PR #1293, opened 2023-05-17
Adding pre-lsdir sync Fs event

sameh-farouk:
Adding to the hooks system new `pre-lsdir` as mandatory Sync Fs Event to enable web-hook or external script execution before reading directory content from Fs.


sameh-farouk:
Accidentally opened for upstream instead of my fork

PR #1332, opened 2023-06-20
check for multiple inverse matches

mistadave:
## Current behaviour
The check `checkEventConditionPatterns` returns after first match is true, which doesn't allow to check multiple wildcards inverse. 

## New behaviour
This change allows in the event rules to set multiple path filters with **inverse matches**. 
Which allows for matching to any kind of file except for those inverse matches (acts as an exclusions list).
image
All tests on the eventmanager_test.go passed.

## Known limitations
Probably doesn't work with mixed path filters. Which I think does not make sense anyway, because a matching pattern always exclude any other inverse matches.

## Changes

* update function checkEventConditionPatterns for multiple inverse patterns
  * Use the _matches_ variable to keep track of the state if condition matches pattern. 
* extend TestDoubleStarMatching with inverse flag
* write test for multiple douple start matches, also for  inverse matches

### DIFF
--- a/internal/common/eventmanager.go
+++ b/internal/common/eventmanager.go
@@ -1159,7 +1159,7 @@ func checkEventConditionPatterns(name string, patterns []dataprovider.ConditionP
 				return false
 			}
 		} else if checkEventConditionPattern(p, name) {
-			matches = true
+			return true
 		}
 	}
 	return matches
--- a/internal/common/eventmanager.go
+++ b/internal/common/eventmanager.go
@@ -1170,10 +1170,8 @@ func checkEventGroupConditionPatters(groups []sdk.GroupMapping, patterns []datap
 		return true
 	}
 	for _, group := range groups {
-		for _, p := range patterns {
-			if checkEventConditionPattern(p, group.Name) {
-				return true
-			}
+		if checkEventConditionPatterns(group.Name, patterns) {
+			return true
 		}
 	}
 
--- a/internal/common/eventmanager.go
+++ b/internal/common/eventmanager.go
@@ -1149,13 +1149,20 @@ func checkEventConditionPatterns(name string, patterns []dataprovider.ConditionP
 	if len(patterns) == 0 {
 		return true
 	}
+	matches := false
 	for _, p := range patterns {
-		if checkEventConditionPattern(p, name) {
-			return true
+		// assume, that multiple InverseMatches are set
+		if p.InverseMatch {
+			if checkEventConditionPattern(p, name) {
+				matches = true
+			} else {
+				return false
+			}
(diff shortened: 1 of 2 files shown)


drakkan:
Hello, thank you for your contribution. I'll give a look the next weekend. Can you please fix the DCO check in the meantime?

internal/common/eventmanager.go
```diff
@@ -1149,13 +1149,20 @@ func checkEventConditionPatterns(name string, patterns []dataprovider.ConditionP
...
-			return true
+		// assume, that multiple InverseMatches are set
+		if p.InverseMatch {
+			if checkEventConditionPattern(p, name) {
+				matches = true
+			} else {
+				return false
+			}
+		} else if checkEventConditionPattern(p, name) {
+			matches = true
```
drakkan, line 1162: Can we just `return true` here?

drakkan:
This patch can be merged but please apply the same logic also in `checkEventGroupConditionPatters`

https://github.com/drakkan/sftpgo/blob/76964a6b8519861d44d435997d469ffc7ccdb985/internal/common/eventmanager.go#L1161

to have uniform behavior. Thank you!

internal/common/eventmanager.go
mistadave: yes that possible, will update that

mistadave:
Hello drakkan,
I've added direct return and added same behaviour to checkEventGroupConditionPatterns(). Did also find a spellmistake.

drakkan:
> Hello drakkan, I've added direct return and added same behaviour to checkEventGroupConditionPatterns(). Did also find a spellmistake.

Thank you, but please fix the [DCO check](https://github.com/drakkan/sftpgo/pull/1332/checks). I'll do some tests with groups this weekend to make sure the behavior is uniform.

mistadave:
Yes, just saw it. Will fix them.

drakkan:
checkEventGroupConditionPatterns seems not consistent, for example if you have something like this

```
Patterns := []dataprovider.ConditionPattern{
		{
			Pattern:      "group1",
			InverseMatch: true,
		},
		{
			Pattern:      "group2",
			InverseMatch: true,
		},
	}

groups = []sdk.GroupMapping{
		{
			Name: "group1",
			Type: sdk.GroupTypePrimary,
		},
		{
			Name: "group4",
			Type: sdk.GroupTypePrimary,
		},
	}

```

I think the check should return `false` because groups must not contains `group1` and `group4`, with your changes it returns `true`. I'll try to fix myself

drakkan:
Hello,

https://cla-assistant.io/drakkan/sftpgo?pullRequest=1332

Thank you!

PR #1359, opened 2023-07-17
Expand reference to http clients in hooks docs

weirdan:
Before, it wasn't really clear where clients were configured and that their configuration also affected headers.


drakkan:
Thank you. Please search for "The HTTP hook will use the global configuration for HTTP clients and will respect the retry" and replace everywhere.

drakkan:
Thank you!

PR #1535, opened 2024-02-09
fix using rsync if running sftpgo as non-root user

jkmw:
If sftpgo is already running as a non-root user and the UID/GID matches the current one, we can skip the syscall.


drakkan:

drakkan:
Thank you!

PR #1538, opened 2024-02-12
[Feature] allow multiple directory mappings (portable mode)

jkmw:
From our perspective, it would be a useful addition to the portable mode if a user could have multiple (virtual) directories.

This PR enables the representation of multiple host paths in the user's directory by specifying one (or more) `-v <host-path>[:<virtual-path>[:<permission>[;<permission>]]]` for the sftp user.


drakkan:
And what if the want to use non-local virtual folders? If you want to use something more complex I suggest using `sftpgo serve` with the memory provider and loading your data from a dump

PR #1548, opened 2024-02-26
SMTP: avoid loading templates if not enabled

crabique:
Hi! 

This small fix aligns SMTP with the current `httpd` behaviour: at the moment, even if `smtp.host` is `''`, it still tries to load the templates before actually disabling SMTP capabilities.

This change allows to `serve` without the templates/bundle if both `httpd` and `smtp` are explicitly disabled.

### DIFF
--- a/internal/smtp/smtp.go
+++ b/internal/smtp/smtp.go
@@ -248,7 +248,7 @@ func (c *Config) loadTemplates(configDir string) error {
 	return nil
 }
 
-// Initialize initialized and validates the SMTP configuration
+// Initialize initializes and validates the SMTP configuration
 func (c *Config) Initialize(configDir string, isService bool) error {
 	if !isService && c.Host == "" {
 		if err := loadConfigFromProvider(); err != nil {
@@ -259,11 +259,11 @@ func (c *Config) Initialize(configDir string, isService bool) error {
 		}
 		return c.loadTemplates(configDir)
 	}
-	if err := c.loadTemplates(configDir); err != nil {
-		return err
-	}
 	if c.Host == "" {
 		return loadConfigFromProvider()
+	}
+	if err := c.loadTemplates(configDir); err != nil {
+		return err
 	}
 	if err := c.validate(); err != nil {
 		return err


drakkan:

crabique:
CLA signed, thanks for the heads-up!

drakkan:
> CLA signed, thanks for the heads-up!

Thanks for signing. I see you have used an email address like users.noreply.github.com. I think this is fine because we just need to make sure that you are entitled to provide a contribution, anyway I asked our lawyers to be sure (this is something I need to know in any case if it happens again in the future). They may be a little slow to respond, please be patient. 

Meanwhile I'll start taking a look at your PR. It looks good at first glance. Thank you!

drakkan:
If you configure SMTP settings from the WebAdmin and try to reset a password SFTPGo will crash after this patch because templates are never loaded in this case.

SMTP can be enabled from WebAdmin, so even if you start without configuration, email templates can be requested at runtime.
To add this functionality we would have to track the loading of the templates and load them when necessary, this would complicate the code a bit and so it will increase the maintenance burden.

Please explain your use case. If you want to execute SFTPGo as a single binary you build it embedding the templates and static resources.

crabique:
Hello! Sorry for the delay, didn't have much time for personal stuff.

Thanks for checking this, I did not know it was possible to enable the feature at runtime from WebAdmin as I never actually used WebAdmin.

My use case was that I just extracted a built binary from a tarball and wanted to spin up a minimal basic SFTP server without any bells and whistles, as I simply don't need anything besides the core SFTP file server. But my minimal config kept failing due to the missing templates, and even having the bundled templates feels wrong if they are not going to be ever used.

A possible way to handle this would be to also check if WebAdmin is enabled, and if it's not, then there will be no way to re-enable SMTP at runtime, so the template loading could be safely skipped.

But I understand it's not the cleanest way and my Go skill / SFTPGo architecture understanding is not good enough to offer a better fix at this time, so I'm completely fine if you close this PR, I tried and I failed 😄

PR #1953, opened 2025-04-18
A11y improvements

eirslett:
A few a11y bugfixes and tweaks.

### DIFF
--- a/static/locales/en/translation.json
+++ b/static/locales/en/translation.json
@@ -103,6 +103,7 @@
         "link": "Go to {{link}}"
     },
     "theme": {
+        "switcher": "Switch theme",
         "light": "Light",
         "dark": "Dark",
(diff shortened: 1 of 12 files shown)


alekssamos:
Hi, @eirslett  . I'm completely blind and was just about to do this. It's a shame the changes haven't been adopted a year later. But it's good to know I'm not the only one.

drakkan:
@alekssamos  we definitely want to support this use case, and we agree it's an important accessibility improvement.
The changes in the PR alone aren't sufficient, we should also update the theme itself (Keenthemes currently don't support accessibility). We haven't forgotten about it, but it needs a more complete implementation than what this PR provides. It's still something we'd like to support in the future, but we don't have a definite timeline for it at the moment. Sorry about that.

PR #1986, opened 2025-05-23
Init script for sysv-init based systems. (#1)

no-systemd:
Init script for sysv-init based systems.

# Checklist for Pull Requests

- [ ] Have you signed the [Contributor License Agreement](https://sftpgo.com/cla.html)?

---

### DIFF
--- a/sftpgo
+++ b/sftpgo
@@ -1,0 +1,102 @@
+#! /bin/sh
+
+### BEGIN INIT INFO
+# Provides:             SFTPGo
+# Required-Start:       $remote_fs $syslog
+# Required-Stop:        $remote_fs $syslog
+# Default-Start:        2 3 4 5
+# Default-Stop:
+# Short-Description:    SFTPGo server
+### END INIT INFO
+
+set -e
+
+# /etc/init.d/sftpgo: start and stop the SFTPGo "server" daemon
+
+SFTPGO_USER="sftpgo"
+SFTPGO_GROUP="sftpgo"
+SFTPGO_BIN_NAME="sftpgo"
+SFTPGO_BIN="/usr/bin/sftpgo"
+SFTPGO_PID="/var/lib/sftpgo/sftpgo.pid"
+SFTPGO_CONF_DIR="/etc/sftpgo"
+SFTPGO_CONF_FILE="sftpgo.json"
+SFTPGO_OPTS="serve -c $SFTPGO_CONF_DIR --config-file $SFTPGO_CONF_FILE"
+
+umask 022
+
+test -x $SFTPGO_BIN || exit 0
+
+
+if test -f /etc/default/$SFTPGO_BIN_NAME; then
+    . /etc/default/$SFTPGO_BIN_NAME
+fi
+
+. /lib/lsb/init-functions
+
+if [ -n "$2" ]; then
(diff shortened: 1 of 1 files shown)


drakkan:
Thanks for the contribution. I believe this works on Debian-based distributions, but—even though I haven't tested it—I'm fairly certain it wouldn't work on something like Slackware. 

We can also merge the PR as-is and address any issues with non-systemd-based distributions if they arise in future contributions. However, please move the script to the [/init](https://github.com/drakkan/sftpgo/tree/main/init) directory. Thank you

no-systemd:
I guess I had an issue with placing the pid file in "/run".
But since "/var/lib/sftpgo" has the correct permissions, and since that is done by either the debian sftpgo package or by sftpgo binary itself, i just placed everything there, to not have the init script doing any of that. (creating files/directories, setting permissions ... etc).
But i understand there is that LSB. So if you wanted everything to go to its LSB place, i could work on that! It is a minmalist script just to get up and running on sysvinit.
So let me please know!
Have a great day! 

On July 3, 2025 2:14:19 PM UTC, Nicola Murino ***@***.***> wrote:
>@drakkan commented on this pull request.
>
>
>
>> +# Required-Start:       $remote_fs $syslog
>+# Required-Stop:        $remote_fs $syslog
>+# Default-Start:        2 3 4 5
>+# Default-Stop:
>+# Short-Description:    SFTPGo server
>+### END INIT INFO
>+
>+set -e
>+
>+# /etc/init.d/sftpgo: start and stop the SFTPGo "server" daemon
>+
>+SFTPGO_USER="sftpgo"
>+SFTPGO_GROUP="sftpgo"
>+SFTPGO_BIN_NAME="sftpgo"
>+SFTPGO_BIN="/usr/bin/sftpgo"
>+SFTPGO_PID="/var/lib/sftpgo/sftpgo.pid"
>
>I think this should be `/run/sftpgo.pid`
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/drakkan/sftpgo/pull/1986#pullrequestreview-2983366007
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>
S. Behnam

drakkan:
@no-systemd It should be fine as it was merged. I have tested in on Devuan, see here:

```
root@devuan:~# /etc/init.d/sftpgo start
Starting SFTPGo server: sftpgo.
root@devuan:~# /etc/init.d/sftpgo status
sftpgo is running.
root@devuan:~# ls -la /run/sftpgo.pid 
-rw-r--r-- 1 root root 5 Jul  4 11:12 /run/sftpgo.pid
root@devuan:~# ps -ef | grep sftpgo
sftpgo    2380     1  0 11:12 ?        00:00:00 /usr/bin/sftpgo serve -c /etc/sftpgo --config-file sftpgo.json
root      2394  2301  0 11:12 pts/1    00:00:00 grep sftpgo
root@devuan:~# /etc/init.d/sftpgo stop
Stopping SFTPGo server: sftpgo.
root@devuan:~# /etc/init.d/sftpgo status
sftpgo is not running ... failed!
```

The init script runs as the root user, so the PID file can be written to /run, which is a tmpfs that is automatically cleaned up after a reboot.

Let me know if you find any issues and thanks for the contribution!

PR #2037, opened 2025-08-05
Add subdirectory option to virtual folders

tecosaur:
Hello!

Thanks for all the work on SFTPGo, it's been difficult to find a good hobbyist solution for exposing some file storage on a NAS with my family. I've read your comments on the difficulty in getting fairly compensated, and I hope the open core model goes well!

### Preamble

In my own usage, I've been trying to set up new-user creation with virtual folders to access content managed by other applications (photos, document uploads, etc.), I've found the inability to specify sub-directories of a virtual folder that should be mapped to a user's tree to be quite inconvenient.

Since this is just a small extension to the filesystem code and data model, I felt like I could give implementing this a shot myself with a bit of help from `grep` and LLMs to make sure I don't miss anything in a codebase I'm unfamiliar with. This seems to have gone fairly well, and I've tested the database upgrade + functionality on my NAS.

### Feature description

Allow virtual folders to map to subdirectories within shared storage, enabling better user isolation and multi-tenant setups.

When configuring a virtual folder, users can now specify a 'virtual_subdirectory' that automatically isolates their access within a shared storage location. For example, multiple users can share the same base folder while each accessing their own subdirectory.

### New UI demo

image

### Checklist for Pull Requests

### DIFF
--- a/examples/ldapauthserver/httpd/models.go
+++ b/examples/ldapauthserver/httpd/models.go
@@ -58,8 +58,9 @@ type SFTPGoFilesystem struct {
 }
 
 type virtualFolder struct {
-	VirtualPath string `json:"virtual_path"`
-	MappedPath  string `json:"mapped_path"`
+	VirtualPath        string `json:"virtual_path"`
+	MappedSubdirectory string `json:"mapped_subdirectory,omitempty"`
+	MappedPath         string `json:"mapped_path"`
 }
 
 // SFTPGoUser defines an SFTPGo user
--- a/examples/php-activedirectory-http-server/README.md
+++ b/examples/php-activedirectory-http-server/README.md
@@ -49,6 +49,18 @@ $virtual_folders['example'] = [
       "quota_files" => -1
     ]
 ];
+
+// Alternative example using mapped_subdirectory for multi-tenant shared storage:
+$virtual_folders['shared_example'] = [
+    [
+      "name" => "shared-documents",
+      "mapped_path" => 'F:\files\shared\documents',
+      "virtual_path" => "/documents",
+      "mapped_subdirectory" => "#USERNAME#",
+      "quota_size" => -1,
+      "quota_files" => -1
+    ]
+];
 ```
 
 ## Example Connection "Output Object" Allowing For No Files in the User's Home Directory ("Root Directory") but Allowing for Files in the Public/Private Virtual Folders
--- a/internal/dataprovider/bolt.go
+++ b/internal/dataprovider/bolt.go
@@ -39,7 +39,7 @@ import (
 )
 
 const (
-	boltDatabaseVersion = 34
+	boltDatabaseVersion = 35
 )
 
 var (
@@ -3170,11 +3170,10 @@ func (p *BoltProvider) migrateDatabase() error {
 		providerLog(logger.LevelError, "%v", err)
 		logger.ErrorToConsole("%v", err)
 		return err
-	case version == 33:
-		logger.InfoToConsole("updating database schema version: %d -> 34", version)
-		providerLog(logger.LevelInfo, "updating database schema version: %d -> 34", version)
-		return updateBoltDatabaseVersion(p.dbHandle, 34)
-
+	case version == 33, version == 34:
+		logger.InfoToConsole("updating database schema version: %d -> 35", version)
+		providerLog(logger.LevelInfo, "updating database schema version: %d -> 35", version)
+		return updateBoltDatabaseVersion(p.dbHandle, 35)
 	default:
 		if version > boltDatabaseVersion {
 			providerLog(logger.LevelError, "database schema version %d is newer than the supported one: %d", version,
@@ -3200,7 +3199,10 @@ func (p *BoltProvider) revertDatabase(targetVersion int) error {
 		logger.InfoToConsole("downgrading database schema version: %d -> 33", dbVersion.Version)
 		providerLog(logger.LevelInfo, "downgrading database schema version: %d -> 33", dbVersion.Version)
 		return updateBoltDatabaseVersion(p.dbHandle, 33)
-
+	case 35:
+		logger.InfoToConsole("downgrading database schema version: %d -> 34", dbVersion.Version)
+		providerLog(logger.LevelInfo, "downgrading database schema version: %d -> 34", dbVersion.Version)
+		return updateBoltDatabaseVersion(p.dbHandle, 34)
 	default:
 		return fmt.Errorf("database schema version not handled: %v", dbVersion.Version)
 	}
--- a/internal/dataprovider/dataprovider.go
+++ b/internal/dataprovider/dataprovider.go
@@ -190,7 +190,10 @@ var (
 	// ErrShareUsageExceeded is returned when reserving share usage tokens would exceed the share max_tokens limit
 	ErrShareUsageExceeded = util.NewI18nError(
 		util.NewRecordNotFoundError("max share usage exceeded"), util.I18nErrorShareUsage)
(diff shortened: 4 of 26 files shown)


drakkan:
Thanks for the contribution! This change needs to be tested carefully, and I don't think I have enough time to review it thoroughly at the moment.
Additionally, it overlaps with the existing group-level placeholders, which already allow you to configure per-user virtual directory paths. For example, if, in your screenshot, `tec` is the username, you can use `%username%` at the group level to achieve the same result

tecosaur:
Thanks for the quick response!

> This change needs to be tested carefully, and I don't think I have enough time to review it thoroughly at the moment.

This sounds completely reasonable to me. I would say that most of the "complexity" of this change is in the breadth, rather than the depth, which should make it somewhat easier at least (glancing at the diff around half of it looks like just adding a `, ""` argument in the appropriate places for when there is no relevant subdir).

I've also paid some attention to the test suite, and have also added a few subdir-specific tests (a few more would probably be good). That said, I think there's a bit more to the testing environment/setup than what I currently have locally, because the FreeBSD CI is showing some issues on tests that aren't run locally by `go test`. I do also see that I overlooked `GetRelativePath` (oops), which I'll push a correction for shortly. 

> Additionally, it overlaps with the existing group-level placeholders, which already allow you to configure per-user virtual directory paths. For example, if, in your screenshot, `tec` is the username, you can use `%username%` at the group level to achieve the same result

I had a look at group-level placeholders before going to all this effort, but they don't quite fit the bill: while I use `tec` as the relevant subdir in the example screenshot, that actually isn't entirely accurate. Since I've just got this working, I am yet to tie in the relevant folders of different applications, but the use of the username/display name/other info for constructing the per-user folders varies by service (e.g. Immich uses display names), and so `%username%` is insufficient.

I think it's valuable to have something more flexible (not just for flexibility's sake, but because it makes it possible to integrate with more services), and not tied to groups.

nickyost:
I think this would be immensely helpful for me for a number of reasons, but namely, only having to connect to each backend once. Right now, each user gets either a Azure Blob Storage backend of one container (or a folder in the container); if I have 10 users or 10 little IoT devices accessing their own data (but not others) thats the same key being pasted into SFTPGo 10 times.

However, if this change is approved and merged into the product, then I can specify one "Virtual Folder" and then assign specific paths to specific users. I can have as many user accounts as I want which all utilize the same Azure Blob Storage connection (Virtual Folder), so when I cycle the key, I only have to cycle one.

This would very likely have commercial value as well for the aforementioned reason. The %username% trick doesnt work for this use case as I may have 3 "users" accessing the same data (maybe even with varying rights) so the folder they access should be the same, but if I use the %username% trick, they don't overlap folders.

Subscribed to the thread. Thanks for the work on SFTPGo as well as thanks to the author of this PR.

tecosaur:
I've gone ahead and resolved the merge conflicts, and while I'm at it, added a bunch of tests (with some help from Claude). This has ~doubled the size of the PR, but it seems worth it for the confidence that this feature is working as expected. It will be interesting to see how the FreeBSD tests go now, given that executes differently to `go test` locally.

At this stage, if anybody has the time, I'd say it's good for review.

tecosaur:
(I forgot to keep the sort order when downgrading the sqlite DB)

tecosaur:
One other thing: given that the `virtual_subdirectory` applies to the mapped path, I'm wondering if the `virtual_` prefix should be stripped?

Dadams2:
This feature would be really useful for me also. I am planning to have multiple users that can setup their own media libraries and grab content from a shared library. There isn't a clean way to do this with existing SFTPgo. 

Being able to specify a virtual subdirectory on mass storage would make this usecase possible. 

I definitely think its worth investing the time to integrate this feature.

tecosaur:
@drakkan I do appreciate that this is a large PR, but if you could be gracious enough to give some indication as to whether/when you might be able to take a look at it, that would be very helpful.

While, as you say `%username%` can be enough for simple situations, as I described in my comment this does not suffice when trying to bridge SFTPGo with other services that use different folder structures.

@nickyost and @Dadams2's comments reinforce my impression that this is a feature that would be generally valuable to have.

I'm willing to resolve the merge conflicts (and there are some other contributions that I am interested in making), but only if there is a real hope of them being reviewed and merged.

drakkan:
@tecosaur To merge this safely, I need to perform a detailed analysis (approx. 1-2 days) to fully understand the architectural implications and potential side effects (e.g. quota management).

Please bear in mind that once a PR is merged, the long-term maintenance falls on the project team, even if the original contributor eventually moves on. Therefore, I need to be completely sure about the implementation.

Right now, we have to prioritize pending requests from enterprise customers, so we cannot give this the focused attention it needs immediately. I will get back to this as soon as my current bandwidth allows. Thanks for understanding

tecosaur:
Thanks for laying out the expected time required to consider this PR properly, and for saying that you do plan on looking at this once there are less pressing requests from paying customers.

Letting me know that you're still interested in looking at this and what's higher on the todo list helps me calibrate my expectations around this PR.

I very much understand the difficulty in splitting attention across an open-source project. Best of luck with your todo list! :slightly_smiling_face:

drakkan:
Thanks for your understanding.

To be transparent about the timeline, my schedule is fully booked for at least the next month and a half with new implementations for the Enterprise version. The good news is that this work also benefits the Community edition too, as fixes and improvements to the shared core are backported (for example, see this [fix](https://github.com/drakkan/sftpgo/commit/f90fbccb2cd98cf38949f4b327ce1adb9b37d238) from yesterday).

That said, looking at your screenshot, I already noticed a potential logic issue even without diving into the code: you are setting a subpath for a virtual folder along with a quota. Currently, quotas apply to the entire virtual folder, not just a subpath.

tecosaur:
Hey Nicola, I've gone ahead and rebased the feature to sit on top of the ~200 new commits over the last 6 months. Along the way, while navigating some of the conflicts like with `os.Root`, I've taken the opportunity to clean up/improve the implementation a bit.

It would be great if you could give this a look :slightly_smiling_face:, I still find this a very useful extension to the virtual folders model.

drakkan:
@tecosaur thanks for your patience. I ended up implementing the feature in a different way. Any help with testing would be greatly appreciated.

The referenced commit comes from a backport of a larger feature that was originally implemented in SFTPGo Enterprise. While backporting the relevant changes, I also fixed a few unrelated issues and ended up pushing everything in a single commit. It's not the cleanest history, sorry about that.

In this AI era, I appreciate real users like you.

tecosaur:
Thanks for getting around to this Nicola. This will be a real help for tying together a few particular services I'm self-hosting.

I took a glance at your implementation, and I see that you more nicely packaged the major changes in VFS and avoided needing to thread subpath support through the different filesystem backends. This does look a bit cleaner/nicer than what I tried here, I think this shows the difference in our familiarity with Go and this codebase :sweat_smile:.

I do have one more change that I'd like to PR, but it's a fair bit smaller and I think shouldn't require as much thought as this did.

Oh, and I'm happy to provide some human interaction. As useful as LLMs can be, the torrent of vibe-coded slop and difficulty telling if you're actually talking with a person have been pretty unpleasant developments :pensive:.

drakkan:
> I do have one more change that I'd like to PR, but it's a fair bit smaller and I think shouldn't require as much thought as this did.

please open an issue describing the feature first. 

> Oh, and I'm happy to provide some human interaction

this and the `os.Root` change are significant features that could expose subtle bugs. Any help with testing would be greatly appreciated.

tecosaur:
> please open an issue describing the feature first

Sure, it's more a bugfix/enhancement than feature though this time.

> Any help with testing would be greatly appreciated.

I'll be taking this for a test drive and report back if I run into anything surprising :slightly_smiling_face:

PR #2100, opened 2025-11-06
bugfix: stat s3 common prefix

Numblgw:
# Checklist for Pull Requests

---
**Description**
image
Directories automatically created via the S3 protocol are returned as common prefixes during list queries. However, some S3 implementations do not support HEAD requests for common prefixes, and empty common prefixes are not automatically removed when objects are deleted. In such cases, when performing a stat operation, both the headObject and getStatForDir functions will return errors, making it necessary to rely on the hasContents function to correctly identify directories (common prefixes). However, when a directory (common prefix) contains only subdirectories (also common prefixes) and no objects, this function fails to work correctly. As a result, the directory (common prefix) cannot be properly removed (rm) or accessed (cd).

### DIFF
--- a/internal/vfs/s3fs.go
+++ b/internal/vfs/s3fs.go
@@ -786,9 +786,10 @@ func (fs *S3Fs) hasContents(name string) (bool, error) {
 	prefix := fs.getPrefix(name)
 	maxKeys := int32(2)
 	paginator := s3.NewListObjectsV2Paginator(fs.svc, &s3.ListObjectsV2Input{
-		Bucket:  aws.String(fs.config.Bucket),
-		Prefix:  aws.String(prefix),
-		MaxKeys: &maxKeys,
+		Bucket:    aws.String(fs.config.Bucket),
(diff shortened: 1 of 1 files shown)


drakkan:
Hello, thanks for the PR.

I assume this change is meant to improve compatibility with implementations like SeaweedFS.
We'll take it into consideration when adding support for directory buckets.

If SeaweedFS (or your target implementation) behaves like S3 when directory buckets are enabled, it will be supported automatically. Otherwise, it’s the external implementation’s responsibility to align with S3 behavior.

In the long term, we’d like to remove workarounds such as [this](https://github.com/drakkan/sftpgo/blob/59bd46a2272d0b94106e059239bb94cb6ac06b48/internal/vfs/s3fs.go#L187) altogether.

drakkan:
Please refer to the following documentation if you are using SeaweedFS. Otherwise, please provide more details about your S3 implementation.

https://github.com/seaweedfs/seaweedfs/wiki/Amazon-S3-API#empty-folders

PR #2111, opened 2025-11-18
enable running on privileged ports

bykvaadm:
# Checklist for Pull Requests

---

when running docker with --network=host you can't run ftp on 21 port.


drakkan:
this change was already [added](https://github.com/drakkan/sftpgo/commit/8d12872608c91c29ba63fa7159647272b24f4afe) in the past and then reverted. If we do something like a common k8s configuration like this:

```
containers[*].securityContext.capabilities:
              drop:
                - ALL
```

will cause issues even if port < 1024 are not required

bykvaadm:
could you please tell how to run sftpgo on 21th port with network mode=host without rebuilding container?

drakkan:
I’m afraid we can’t support this setup without introducing an additional container variant, and we’re not in a position to maintain that at the moment.

Binding to privileged ports requires extra capabilities, which can cause problems in environments that enforce restricted security contexts (for example when all capabilities are dropped, such as on OpenShift). Supporting both configurations would require maintaining multiple images, which isn’t feasible for us right now.

If you need to run SFTPGo on port 21 with network_mode=host, the recommended approach is to build a custom image based on ours.
Thanks for using SFTPGo

bykvaadm:
> If you need to run SFTPGo on port 21 with network_mode=host, the recommended approach is to build a custom image 
> based on ours.

yes, i'm already doing this.

> I’m afraid we can’t support this setup without introducing an additional container variant, and we’re not in a position to > maintain that at the moment.

How can i help you? we can add Dockerfile.privileged to have another build from maintainer

PR #2065, opened 2025-09-18
Log ciphersuite in ftp calls

GitBluub:
We want to log ciphersuite used by our clients in sftpgo to be able to send deprecation notices and removes unsafe cipher suites currently used.
In order to do this I have to be able to link a client to its net.Conn used. This is a currently working for me, non-mergeable implementation (especially because I had to fork your fork of ftpserverlib to be able to get underlying net.Conn) but also because it might not be the best implementation for this.

it results in login logs like this
```
"level":"info","time":"2025-09-18T09:51:26.741","sender":"login","ip":"X.X.X.X","username":"me","method":"password","protocol":"FTP","connection_id":"FTP_0_2","client":"Unknown","encrypted":true,"ciphersuite":"TLS_AES_128_GCM_SHA256"}
```

Anyway if you agree on an implementation like this it requires another PR in your ftpserverlib fork (the only change was a `Conn` function in ClientContext interface to return underlying net.Conn

# Checklist for Pull Requests

- [ ] Have you signed the [Contributor License Agreement](https://sftpgo.com/cla.html)?

---

PR #2124, opened 2025-11-27
Log ciphersuite in ftp calls

GitBluub:
Hello, I am back with a small PR using the Extra field to store the ciphersuite used in the data connection. I made a small TlsState struct to store login + ciphersuite but hesitated storing the whole tls.ConnectionState struct inside. I think we can just add the field we want if someone needs/wants more like tlsversion or whatever

We want to log ciphersuite used by our clients in sftpgo to be able to send deprecation notices and removes unsafe cipher suites currently used. Late followup of #2065 

# Checklist for Pull Requests

---

### DIFF
--- a/internal/ftpd/server.go
+++ b/internal/ftpd/server.go
@@ -239,16 +239,13 @@ func (s *Server) WrapPassiveListener(listener net.Listener) (net.Listener, error
 
 // VerifyConnection checks whether a user should be authenticated using a client certificate without prompting for a password
 func (s *Server) VerifyConnection(cc ftpserver.ClientContext, user string, tlsConn *tls.Conn) (ftpserver.ClientDriver, error) {
-	if !s.binding.isMutualTLSEnabled() {
-		return nil, nil
-	}
 	if tlsConn != nil {
 		state := tlsConn.ConnectionState()
 		cc.SetExtra(&tlsState{
 			Cipher: tls.CipherSuiteName(state.CipherSuite),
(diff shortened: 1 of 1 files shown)


internal/ftpd/server.go
```diff
@@ -36,6 +36,14 @@ import (
 	"github.com/drakkan/sftpgo/v2/internal/version"
 )
 
+// TlsState tracks TLS connection state for a client
+type TlsState struct {
+	// LoginViaTLS indicates whether the user logged in using TLS certificate authentication
+	LoginViaTLS bool
```
drakkan: please change to mTLS

GitBluub:
Updated and tested, following your review I renamed LoginViaTLS to LoginWithMutualTLS, CipherSuiteName to Cipher, made the struct private and fixed a test for setExtra. The empty `cc.SetExtra(&TlsState{})` was removed as well because it made no sense.

drakkan:
Test with TLS enabled and mutual TLS disabled. Does it work?

GitBluub:
It only worked when client auth-type was different than 0 because of the early return in VerifyConnection. I checked and VerifyConnection is called any time it can by ftpserverlib so I should have moved the mTLS check after the cipher set so that it can work without mTLS enabled.
I edited the Verifyconnection to check for mTLS later and still keep the same behavior. I tested it and now it works with mTLS disabled as well, thanks for the notice

PR #2162, opened 2026-01-22
Add fix for SSH_FXP_STAT with S3 backend during file upload

### LINKED ISSUE #2010: [Bug]: SSH_FXP_STAT fails with S3 backend

### Bug description

Hi,

I have a program that uploads files to an SFTP server and changes its mode bits. This program works with several SFTP servers, but fails with SFTPGo when using an S3 backend (MinIO). The following SFTP requests are sent:

1. SSH_FXP_OPEN to open the file for writing
2. one or more SSH_FXP_WRITE requests to write data into it
3. SSH_FXP_STAT to query the current file attributes for modification
...

However, the SSH_FXP_STAT request fails with a "no such file" error, if the file does not exist before uploading.

### Steps to reproduce

See the description.

### Expected behavior

SSH_FXP_STAT should work for the file.

### SFTPGo version

2.6.6

### Data provider

SQLite

### Installation method

Community Docker image

### Configuration

Added administrative user, normal user and setup a bucket on the MinIO backend (RELEASE.2025-06-13T11-33-47Z) as a virtual folder mapped for the user.

### Relevant log output

```shell

```

Private user, home usecase (home backup/VPS)

### Additional info

_No response_

jstudler:
This fixes an issue with S3 backed virtual folders where the SFTP client tries to retrieve SSH_FXP_STAT while the file is still open for writing. The implemented fix checks whether there is a file upload to S3 in progress and then returns a valid SSH_FXP_STAT response to the SFTP client.

Fixes  #2010 

# Checklist for Pull Requests

# Testing

The following python script was used for testing the behavior. Make sure to configure an S3 backed virtual folder under `mys3virtualfolder` in the users home directory. I used uv to run the test. Also make sure to configure SFTPGo to ignore stat changes (`setstat_mode` set to `1`)

```
go run main.go serve --log-level debug --config-file sftpgo.yaml
uv run test_stat_during_upload.py --host localhost --port 2022 --username <username> --password <password> --path mys3virtualfolder/testfile
```

```yaml
common:
  setstat_mode: 1
```

`test_stat_during_upload.py`: https://gist.github.com/jstudler/123c15740e48637fa40fe3e31ab4a16e

edit 2026-01-22: move test_stat_during_upload.py to gist to reduce size of PR description

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -463,6 +463,7 @@ type ActiveTransfer interface {
 	GetDownloadedSize() int64
 	GetUploadedSize() int64
 	GetVirtualPath() string
+	GetFsPath() string
 	GetStartTime() time.Time
 	SignalClose(err error)
 	Truncate(fsPath string, size int64) (int64, error)
--- a/internal/common/connection.go
+++ b/internal/common/connection.go
@@ -296,6 +296,29 @@ func (c *BaseConnection) setTimes(fsPath string, atime time.Time, mtime time.Tim
 	return false
 }
 
+// getStatForUploadTransfer checks if there's an active upload transfer for the given path.
+// If found, it returns file info based on the current transfer state.
+// This is useful for S3-backed storage where the file doesn't exist on the backend until the upload completes.
+func (c *BaseConnection) getStatForUploadTransfer(fsPath string) (os.FileInfo, bool) {
+	c.RLock()
+	defer c.RUnlock()
+
+	c.Log(logger.LevelDebug, "checking for active upload transfer, fsPath: %q, active transfers: %d", fsPath, len(c.activeTransfers))
+	for _, t := range c.activeTransfers {
+		c.Log(logger.LevelDebug, "  transfer id: %d, type: %d (upload=%d), fsPath: %q, match: %v",
+			t.GetID(), t.GetType(), TransferUpload, t.GetFsPath(), t.GetFsPath() == fsPath)
+		if t.GetType() == TransferUpload && t.GetFsPath() == fsPath {
+			// Return file info with current upload size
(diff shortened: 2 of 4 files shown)


drakkan:
Thanks for the proof-of-concept PR. One clear issue is that this approach won’t work in multi-instance installations, which would introduce inconsistent behavior between single-instance and clustered setups.

Adding a fallback to support cluster installations could also be quite heavy, especially if it is triggered on every “file not found” error.

I’ll take a more in-depth look when I have some time. It’s also worth noting that the current behavior is consistent with the underlying storage backend, which only makes files visible once the upload is completed.

jstudler:
Thanks for your feedback and for taking it into consideration. It would be extremely helpful for some of our use cases to allow this to be implemented in SFTPGo. Unfortunately we are not in full control of all the SFTP client implementations and the only way to fix this is on the SFTP server side.

I agree that depending on the implementation of clustering it can be challenging to make this consistent across instances.

An option could be to create a new configuration parameter `getstat_mode` so this behavior can be enabled by the SFTPGo admin. This would also allow to properly document the behaviour in a clustered setup. Alternatively, I could also imagine adding it as a new boolean configuration parameter for S3 backed virtual folders so the setting is per virtual folder and where explicitly needed.

drakkan:
I think this PR doesn’t work the way you expect, and that’s actually good news. `c.activeTransfers` returns the transfers associated with the current active connection (typically just one), not all active transfers. Because of that, there’s no expectation that this would work across a different connection or a different cluster instance. This is the same logic we already use for chtimes: an ongoing transfer is only visible from the connection it originated from. This way is acceptable

jstudler:
Thank you very much @drakkan for the implementation and merge!

PR #2164, opened 2026-01-26
Fix/ftp active connection closed

mmtootmm:
Thanks for maintaining this great project.

# Checklist for Pull Requests

---

## Problem
FTP logins occasionally fail because fresh, unauthenticated connections are prematurely closed by the idle connection check, regardless of the configured timeouts.

## Cause
When a new FTP connection is initialized, it uses an empty `dataprovider.User{}` struct:
https://github.com/drakkan/sftpgo/blob/e3b2780655c3f8a38649e0c5f3fc8f6c50fe12ad/internal/ftpd/server.go#L170-L176

Since an empty `User` struct has a default status (which is not "Enabled"), the `c.isAccessAllowed()` check returns `false`:
https://github.com/drakkan/sftpgo/blob/e3b2780655c3f8a38649e0c5f3fc8f6c50fe12ad/internal/common/connection.go#L115-L120

Consequently, these connections are flagged and closed in the idle connection loop:
https://github.com/drakkan/sftpgo/blob/e3b2780655c3f8a38649e0c5f3fc8f6c50fe12ad/internal/common/common.go#L1164-L1170

## What I did
- **Updated Logic:** Bypassed the `!c.isAccessAllowed()` check if the connection belongs to an unauthenticated FTP user.
- **Improved Test Case:** Updated the test case to use `dataprovider.User{}` instead of a User with `Status: 1`, ensuring the test accurately reflects real-world scenarios.

## Alternative Consideration
Another solution could be to initialize new FTP connections with a User struct where `Status` is explicitly set to `1` (Enabled). I'm open to discussing which approach fits the project's architecture better.

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -1161,7 +1161,7 @@ func (conns *ActiveConnections) checkIdles() {
 				logger.Debug(conn.GetProtocol(), conn.GetID(), "close idle connection, idle time: %s, username: %q close err: %v",
 					time.Since(conn.GetLastActivity()), conn.GetUsername(), err)
 			}(c)
-		} else if !c.isAccessAllowed() {
+		} else if !isUnauthenticatedFTPUser && !c.isAccessAllowed() {
 			defer func(conn ActiveConnection) {
 				err := conn.Disconnect()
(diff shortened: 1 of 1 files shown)


drakkan:
thank you

PR #2170, opened 2026-02-03
Support LATIN1 encoding on FTP paths

rdavaillaud:
Possible fix to the LATIN1 encoding support for FTP server (https://github.com/drakkan/sftpgo/discussions/2169).

Disclaimer: I asked Codex to do the work, I am not a Go dev.

# Checklist for Pull Requests

---

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -67,6 +67,7 @@ require (
 	golang.org/x/oauth2 v0.34.0
 	golang.org/x/sys v0.39.0
 	golang.org/x/term v0.38.0
+	golang.org/x/text v0.32.0
 	golang.org/x/time v0.14.0
 	google.golang.org/api v0.257.0
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
@@ -169,7 +170,6 @@ require (
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
 	golang.org/x/mod v0.31.0 // indirect
 	golang.org/x/sync v0.19.0 // indirect
-	golang.org/x/text v0.32.0 // indirect
 	golang.org/x/tools v0.40.0 // indirect
 	golang.org/x/xerrors v0.0.0-20240903120638-7835f813f4da // indirect
 	google.golang.org/genproto v0.0.0-20251222181119-0a764e51fe1b // indirect
--- a/internal/ftpd/handler.go
+++ b/internal/ftpd/handler.go
@@ -22,9 +22,11 @@ import (
 	"path"
 	"strings"
 	"time"
+	"unicode/utf8"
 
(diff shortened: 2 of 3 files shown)


drakkan:
Thanks for the PR.

I have significant architectural concerns regarding the introduction of non-UTF-8 character sets into the core logic:

- Hardcoding ISO-8859-1 creates an arbitrary bias. It solves the issue for Western European legacy systems but leaves users requiring GBK, Shift_JIS, or Windows-1251 (Cyrillic) with no solution. A proper implementation would require a configurable charset decoder/encoder interface using golang.org/x/text/encoding, not a hardcoded fallback.
- SFTPGo relies on Go's string type, which assumes UTF-8. More importantly, SFTPGo abstracts multiple storage backends (S3, GCS, Azure). Cloud providers strictly require valid UTF-8 for object keys. Passing raw non-UTF-8 bytes (converted from Latin1) to these SDKs could result in API errors, unreachable objects, or data corruption in the backing storage. Ensuring data integrity across all backends with non-standard encodings introduces a massive testing surface.
- This approach breaks interoperability in a mixed environment. If a modern client uploads a file with characters outside the ISO-8859-1 range (e.g., Emojis, Chinese, or Greek characters), the server cannot represent these filenames to the legacy client. The encoding would likely fallback to replacement characters (e.g., file_??.txt). If the legacy client tries to download file_??.txt, the operation will fail with a "File not found" error because the actual filename on storage is different. This creates a broken state where files are visible but inaccessible.
- Modern FTP demands UTF-8 (RFC 2640, published in 1999). Implementing a fallback mechanism encourages non-compliance.

At this stage, my assessment is that the maintenance burden and architectural complexity of supporting non-UTF-8 encodings (correctly and safely across all backends) outweigh the benefits for the few legacy use cases remaining.

You are, of course, welcome to build a custom version of SFTPGo with this patch for your specific needs. I might be open to revisiting this in the future if there is a proven history of stability in production, but currently, I cannot commit the resources required to test and maintain this feature in the main codebase

rdavaillaud:
@drakkan Hi, 
Thanks to you for your review.
I totally understand your point, and obviously, the "simple" modification in this PR were not enough. 

> Modern FTP demands UTF-8 (RFC 2640, published in 1999). Implementing a fallback mechanism encourages non-compliance.

This is where my problem is. Unfortunately, lot of the software that calls my ftp did not upgrade their clients.
Anyway, for now, I've managed to do a somehow complicated but working workaround with a FTP Proxy and a "transcoding" middleware. Not perfect, not really stable, but the job is done... mostly 😅 

When I have time, I would try to build something out of all this. 
Thank you

Issue #2157, opened 2026-01-11
[Feature Request] proxy_allowed allowing hostname input


TheLonelinessOfHS:
### Is your feature request related to a problem? Please describe.

Scenario: I use docker swarm which doesn't allow me to set a static ip for my traefik reverse proxy container. To make SFTP work, I had to set a IP range for proxy_allowed. For now, the best practice is to use hostname (e.g., tasks.traefik) to identify the requests forwarded by the reverse proxy. I am wondering if you could have a look at this? Thanks.

### Describe the solution you'd like

Allow proxy_allowed to take hostnames

### What are you using SFTPGo for?

Private user, home usecase (home backup/VPS)

PR #2204, opened 2026-04-04
Feat(): Add possibility to insert hostname in proxy_allowed

ghost:
# Checklist for Pull Requests

This commit add the possibility to use hostname in proxy_allowed variable in sftpgo.json config file

---

### DIFF
--- a/internal/common/common_test.go
+++ b/internal/common/common_test.go
@@ -249,14 +249,17 @@ func TestInitializationProxyErrors(t *testing.T) {
 
 	c := Configuration{
 		ProxyProtocol: 1,
-		ProxyAllowed:  []string{"1.1.1.1111"},
+		// "not valid ip" contains spaces so it is rejected as neither a valid IP,
+		// CIDR, nor hostname.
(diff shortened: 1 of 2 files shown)


blaireauvpn:
Hello, i would like to contribute to the projet to add a feature i wanted when combining sftpgo + docker swarm or kubernetes, i also saw an open issue #2157  about it in the issue section.

Regards, Jérôme

drakkan:
Thanks for the proposal. I'd prefer not to introduce runtime DNS resolution in proxy_allowed: it adds latency in the connection path, can block on slow/failing DNS and keeps stale entries on lookup errors
   
For the Docker Swarm use case, you can attach Traefik and SFTPGo to a dedicated overlay network with a fixed subnet, e.g.:

```
networks:
  sftpgo_net:
    driver: overlay
    ipam:    
      config:   
        - subnet: 10.20.0.0/24
```
            
Then set proxy allowed to `["10.20.0.0/24"]`. Container IPs stay dynamic, but they're always drawn from that range, no DNS needed, and the check stays a simple CIDR match.

PR #2210, opened 2026-04-10
feat: enable multi-language by default via configuration file

o-i-i-o:
feat: enable multi-language by default via configuration file

### DIFF
--- a/sftpgo.json
+++ b/sftpgo.json
@@ -285,7 +285,12 @@
         "render_openapi": true,
         "base_url": "",
         "languages": [
-          "en"
+          "de",
+          "en",
+          "es",
+          "fr",
+          "it",
+          "zh-CN"
         ],
         "oidc": {
           "client_id": "",
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -121,7 +121,7 @@ var (
 		HideLoginURL:         0,
 		RenderOpenAPI:        true,
 		BaseURL:              "",
-		Languages:            []string{"en"},
+		Languages:            []string{"de", "en", "es", "fr", "it", "zh-CN"},
 		OIDC: httpd.OIDC{
 			ClientID:                   "",
 			ClientSecret:               "",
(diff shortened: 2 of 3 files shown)


sftpgo.json
```diff
@@ -285,7 +285,12 @@
         "render_openapi": true,
         "base_url": "",
         "languages": [
-          "en"
+          "de",
```
utafrali, line 288: This updates the example/default config file, but the in-code default for `Binding.Languages` (internal/httpd/httpd.go) still appears to be `["en"]`. When a user runs `sftpgo gen config` to generate a fresh config, they will get the old English-only default rather than the new multi-language set defined here. The Go-level default should be updated to match, e.g.:

```go
Languages: []string{"de", "en", "es", "fr", "it", "zh-CN"},
```

Otherwise `sftpgo.json` and the generated config will diverge, which is confusing for new users.

utafrali, line 288: Only `bindings[0]` is updated here. If the default config ever gains a second binding entry (e.g. a TLS binding on 8443 that some users copy-paste from docs), it would silently inherit only `["en"]`. This is fine for the current single-binding default, but worth noting that each binding needs its own `languages` array if multiple bindings are in use.

utafrali (commented):
The JSON change itself is correct, well-formed, and all 6 added languages have corresponding translation files in /static/locales/. However, the in-code default for `Binding.Languages` does not appear to have been updated alongside sftpgo.json, meaning `sftpgo gen config` will still produce an English-only config and contradict the intent of this PR.

sftpgo.json
o-i-i-o: I understand that the configuration file is generated at line 124 in ./internal/config/config.go. However, the root sftpgo.json is used during packaging, and it appears that sftpgo.json is not generated from Go code. We know that all the translation files are complete, but there may be other reasons why the original author did not enable them directly. I believe it is more reasonable to modify sftpgo.json directly instead of ./internal/config/config.go, as this enables multi-language support without changing the Go code.

PR #2227, opened 2026-05-04
feat: enable shared mode for sqlite

azamaulanaaa:
SQLite3 has active-passive replication via litefs thus allowing us to use it in multi-instances of services. since table for share session already in the code of sqlite, why not enable it?

# Checklist for Pull Requests

- [ x] Have you signed the [Contributor License Agreement](https://sftpgo.com/cla.html)?

---


drakkan:
> SQLite3 has active-passive

SFTPGo shared mode requires an active-active setup

PR #2231, opened 2026-05-09
fix: compare duplicate slice values correctly

immanuwell:
## What
`util.SlicesEqual` now checks duplicate counts, not only membership.

Small footgun, but it can matter for admin self-permission checks.

## Repro
Before this patch:

`SlicesEqual([]string{"read", "write"}, []string{"read", "read"}) == true`

same length, different values, still true. not it tbh.

## Tests
`go test ./internal/util -count=1`

`go test -tags nopgxregisterdefaulttypes,disable_grpc_modules -p 1 -timeout 5m ./internal/httpd -run 'TestAdminUpdateSelf|TestAdminUpdateSelfMock' -count=1`

## Checklist

### DIFF
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -44,7 +44,6 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
-	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


drakkan:
merged with minor modifications. Thank you

PR #2250, opened 2026-06-18
Harden local filesystem writes against symlink escape

fabwa:
## Summary

- validate local OS filesystem write targets before create and rename operations
- preserve configured atomic upload temp-file behavior while revalidating the final write target
- add regression tests for dangling symlink writes, safe in-root symlink writes, atomic temp files, and symlinked rename parents

## Security implication

On local filesystem backends, write and atomic-rename operations should not follow symlinks or stale parent paths outside the configured filesystem root. Without this extra validation, an authenticated user with file-write and symlink-capable permissions could write as the SFTPGo process user outside their authorized storage boundary. Depending on deployment permissions, that can allow modification of trusted files and may lead to code execution if a writable hook, plugin, config, service, template, or deployment path is targeted.

## CWEs

- CWE-59: Improper Link Resolution Before File Access
- CWE-22: Improper Limitation of a Pathname to a Restricted Directory
- CWE-73: External Control of File Name or Path
- CWE-367: Time-of-check Time-of-use Race Condition

## Tests

- `go test ./internal/vfs`
- `go test ./internal/sftpd -run 'Test(Link|EscapeHomeDir|ResolvePaths)$' -count=1`

### DIFF
--- a/internal/vfs/osfs.go
+++ b/internal/vfs/osfs.go
@@ -134,6 +134,11 @@ func (fs *OsFs) Open(name string, offset int64) (File, PipeReader, func(), error
 
 // Create creates or opens the named file for writing
 func (fs *OsFs) Create(name string, flag, _ int) (File, PipeWriter, func(), error) {
+	if !fs.isAtomicUploadTempPath(name) {
+		if err := fs.validateWriteTarget(name); err != nil {
+			return nil, nil, nil, err
+		}
+	}
 	if !fs.useWriteBuffering(flag) {
 		var err error
 		var f *os.File
@@ -178,6 +183,9 @@ func (fs *OsFs) Create(name string, flag, _ int) (File, PipeWriter, func(), erro
 func (fs *OsFs) Rename(source, target string, checks int) (int, int64, error) {
 	if source == target {
 		return -1, -1, nil
+	}
+	if err := fs.validateWriteTarget(target); err != nil {
+		return -1, -1, err
 	}
 	err := os.Rename(source, target)
 	if err != nil && isCrossDeviceError(err) {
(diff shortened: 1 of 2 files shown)


drakkan:
If you are interested to help, please test this change:

https://github.com/drakkan/sftpgo/commit/661d77f47b8df04f44c5e7e4357d57e27b763743

Thank you

fabwa:
From my side this looks sufficient for the original issue I was trying to address. Moving `OsFs`/`CryptFs` to `os.Root` is a stronger fix than my pre-check approach because the final create/open/rename operations now happen through the confined root handle instead of relying only on path validation before the syscall.

There are a couple of residual areas I think are worth tracking separately, not as blockers for this fix:

Go documents that os.Root does not prevent traversal across filesystem boundaries such as bind mounts, /proc, or device files placed under the root. That is probably an admin/threat-model constraint, but it may be worth documenting for local filesystem backends.

Root.Chmod, Root.Chown, and Root.Chtimes have documented Unix race caveats. SFTPGo exposes these through SetStat operations, so it may be worth either adding targeted tests, documenting the residual semantics, or using descriptor-based / no-follow variants where possible.

CheckRootPath still uses direct os.MkdirAll before opening the root. That seems fine as provisioning logic, but it is another operation outside the os.Root model and should probably stay explicitly out of the user-controlled path threat model.

My suggestion would be to treat 661d77f as fixing the write/rename symlink escape, then open a separate follow-up issue for “local filesystem residual confinement semantics” covering SetStat metadata operations and mounted special filesystems.

Cheers

drakkan:
The updated doc is not yet live but already pushed to the repo:

https://github.com/sftpgo/docs/commit/ba1f95ea8af6822d9d8f5fbee0fa9c9a6a60902e
https://github.com/sftpgo/docs/commit/702788d5745c470f23437bcf2c834a0886b80691

Is it clear enough?

Root.Chmod/Chown/Chtimes races are only possible within the root, the worst case is the operation applying to the symlink file itself, inside the root. Only someone who can already write inside that root can trigger it, on their own files, so there is no confinement impact.

CheckRootPath and os.MkdirAll: correct, the provisioning is by definition outside the os.Root model. The path comes from the admin configuration not from a client-supplied path so it is already outside the user-controlled threat model.

What would really help is testing this new implementation against real workloads. If you're an SFTPGo user, that's what matters most :smile:

PR #2253, opened 2026-06-26
feat(ftpd): add passive port multiplexing by client IP

flozano:
## Summary

ftpserverlib v0.31.0 added passive port multiplexing by client IP via [fclairamb/ftpserverlib#623](https://github.com/fclairamb/ftpserverlib/pull/623), exposed as the `Settings.PassiveTransferPortMultiplexing` flag. SFTPGo already depends on ftpserverlib v0.32.0 but does not surface the option, so this PR wires it through the FTP configuration.

When enabled, FTP clients with **different** IP addresses can share the same passive port concurrently, instead of requiring one listener per session. This increases the number of simultaneous passive transfers that can be served from a limited passive port range. Clients behind the same NAT / public egress IP are treated as a single client for multiplexing purposes, so the effective limit is based on the observed source IP.

The option is disabled by default, preserving the current behavior.

## Changes

- `internal/ftpd/ftpd.go`: new `PassivePortMultiplexing` field on `Configuration` (`passive_port_multiplexing`), placed next to the existing `passive_port_range` option.
- `internal/ftpd/server.go`: map it to `ftpserver.Settings.PassiveTransferPortMultiplexing` when building the server settings.
- `internal/config/config.go`: register the viper default, alongside the other `ftpd.*` defaults.
- `sftpgo.json`: document the new key in the example configuration.

The naming mirrors SFTPGo's existing convention, which drops `Transfer` from the upstream names (`PassiveTransferPortRange` → `passive_port_range`), giving the parallel `passive_port_multiplexing` key. The exposure follows the exact same pattern as the sibling boolean flags such as `disable_active_mode`.

## Testing

- `go build ./...`
- `go vet ./internal/ftpd/... ./internal/config/...`
- `gofmt` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -2118,6 +2118,7 @@ func setViperDefaults() {
 	viper.SetDefault("ftpd.active_transfers_port_non_20", globalConf.FTPD.ActiveTransfersPortNon20)
 	viper.SetDefault("ftpd.passive_port_range.start", globalConf.FTPD.PassivePortRange.Start)
 	viper.SetDefault("ftpd.passive_port_range.end", globalConf.FTPD.PassivePortRange.End)
+	viper.SetDefault("ftpd.passive_port_multiplexing", globalConf.FTPD.PassivePortMultiplexing)
 	viper.SetDefault("ftpd.disable_active_mode", globalConf.FTPD.DisableActiveMode)
 	viper.SetDefault("ftpd.enable_site", globalConf.FTPD.EnableSite)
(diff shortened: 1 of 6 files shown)


drakkan:
Thanks for the PR, and apologies for the delay. It’s an interesting feature that I think is worth adding; however, after a quick check of the ftpserverlib PR, I believe there are some correctness issues. I need to find the time to review it properly. This is just to let you know that is something I'll try to merge although I don't know when yet.

PR #2255, opened 2026-06-29
vfs: include the container when comparing Azure Blob resources

### LINKED ISSUE #2254: [Bug]: Azure Blob isSameResource ignores the container — cross-container copy/move fails

### Bug description

Copying (or moving) a file between two virtual folders that are backed by **different containers within the same Azure Blob storage account** fails.

The root cause is that `AzBlobFsConfig.isSameResource()` does not take the **container** into account. It compares only account name, endpoint and SAS URL:

```go
// internal/vfs/vfs.go
func (c *AzBlobFsConfig) isSameResource(other AzBlobFsConfig) bool {
	if c.AccountName != other.AccountName {
		return false
	}
	if c.Endpoint != other.Endpoint {
		return false
	}
	if c.SASURL == nil {
		c.SASURL = kms.NewEmptySecret()
	}
	if other.SASURL == nil {
		other.SASURL = kms.NewEmptySecret()
	}
	return c.SASURL.GetPayload() == other.SASURL.GetPayload()
}
```

Because of this, two Azure Blob folders that point at different containers in the same account are treated as the *same resource*. This has two consequences:

1. **Copy of a file with an unchanged name is rejected.** `BaseConnection.checkCopy()` does:

   ```go
   if c.IsSameResource(virtualSource, virtualTarget) {
       if fsSourcePath == fsTargetPath {
           return fmt.Errorf("the copy source and target cannot be the same: %w", c.GetOpUnsupportedError())
       }
   }
   ```

   For two different containers the resolved fs paths are the container-relative blob keys, so `containerA/file.txt` and `containerB/file.txt` both resolve to `file.txt` and are considered equal. The copy is rejected with `ErrOpUnsupported`, surfaced as HTTP 400 in the web client copy task — instantly, with no I/O.

2. **Server-side copy/rename to a different name targets the wrong container.** When the names differ, the `FsFileCopier` fast path is taken and `AzureBlobFs.copyFileInternal()` builds *both* the source and destination block-blob clients from a single `fs.containerClient` (the target folder's container):

   ```go
   srcBlob := fs.containerClient.NewBlockBlobClient(source)
   dstBlob := fs.containerClient.NewBlockBlobClient(target)
   ```

   So the source blob is looked up in the target container, where it does not exist. Cross-container move/rename (`executeRenameTask` → `IsSameResource == true` → server-side `Rename`) is affected the same way.

### Steps to reproduce

1. Configure two virtual folders backed by Azure Blob, same `account_name` and `endpoint`, **different `container`** (e.g. `container-a` and `container-b`), no SAS URL.
2. Upload a file `file.txt` into `container-a`.
3. Via the web client (or SFTP), copy `/folder-a/file.txt` to `/folder-b/file.txt`.
4. The copy task finishes with HTTP 400 (`the copy source and target cannot be the same`). Copying to a different name instead fails to find the source blob.

### Expected behavior

Copying/moving a file between two folders on different Azure containers should succeed (e.g. via the cross-resource streaming copy). `isSameResource` should treat different containers as different resources.

### SFTPGo version

v2.7.3 (also present on `main` as of this writing)

### Data provider

SQLite

### Installation method

Community Docker image

### Configuration

Two Azure Blob virtual folders sharing one storage account, distinguished only by container name, authenticated via account key (no SAS URL). Default data provider.

### Relevant log output

```shell
{"level":"debug","sender":"httpd","method":"POST","uri":"/web/client/file-actions/copy?path=%2Ffolder-a%2Ffile.txt&target=%2Ffolder-b%2Ffile.txt","resp_status":202}
{"level":"debug","sender":"OIDC","message":"copy task finished, status: 400, update task err: <nil>"}
```

Medium business

### Additional info

Suggested fix: add a container comparison to `AzBlobFsConfig.isSameResource()`, and optionally a `KeyPrefix` comparison. That routes cross-container copies/moves through the existing streaming path while preserving the server-side fast path for same-container operations.

philipp-holler:
## Summary

`AzBlobFsConfig.isSameResource()` compares only account name, endpoint and SAS URL, ignoring the container. As a result two Azure Blob virtual folders that point at different containers in the same storage account are treated as the same resource.

This breaks copy/move between such folders:

- A copy that keeps the file name is rejected by `checkCopy` with `the copy source and target cannot be the same` (HTTP 400), because the container-relative paths are equal.
- A copy/move to a different name takes the `FsFileCopier` server-side fast path, where `copyFileInternal` builds both the source and destination block-blob clients from a single `containerClient` (the target's), so the source blob is looked up in the wrong container.

This PR adds a container comparison to `isSameResource`, mirroring `isEqual` and the bucket comparison used for the other object-storage providers. Different containers are now correctly reported as different resources, so cross-container copy/move falls back to the streaming (download + upload) path and succeeds; same-container operations keep the server-side fast path.

Fixes #2254

## Test

Extended `TestVfsSameResource` with an Azure Blob case covering same vs. different container.

### DIFF
--- a/internal/common/common_test.go
+++ b/internal/common/common_test.go
@@ -1655,6 +1655,29 @@ func TestVfsSameResource(t *testing.T) {
 	res = fs.IsSameResource(other)
 	assert.False(t, res)
 	fs = vfs.Filesystem{
+		Provider: sdk.AzureBlobFilesystemProvider,
+		AzBlobConfig: vfs.AzBlobFsConfig{
+			BaseAzBlobFsConfig: sdk.BaseAzBlobFsConfig{
+				AccountName: "a",
+				Container:   "c1",
+			},
+		},
+	}
+	other = vfs.Filesystem{
+		Provider: sdk.AzureBlobFilesystemProvider,
+		AzBlobConfig: vfs.AzBlobFsConfig{
+			BaseAzBlobFsConfig: sdk.BaseAzBlobFsConfig{
+				AccountName: "a",
+				Container:   "c1",
+			},
+		},
+	}
+	res = fs.IsSameResource(other)
+	assert.True(t, res)
+	other.AzBlobConfig.Container = "c2"
+	res = fs.IsSameResource(other)
+	assert.False(t, res)
+	fs = vfs.Filesystem{
 		Provider: sdk.HTTPFilesystemProvider,
 		HTTPConfig: vfs.HTTPFsConfig{
 			BaseHTTPFsConfig: sdk.BaseHTTPFsConfig{
--- a/internal/vfs/vfs.go
+++ b/internal/vfs/vfs.go
@@ -818,6 +818,9 @@ func (c *AzBlobFsConfig) tryDecrypt() error {
 }
 
 func (c *AzBlobFsConfig) isSameResource(other AzBlobFsConfig) bool {
+	if c.Container != other.Container {
+		return false
+	}
 	if c.AccountName != other.AccountName {
 		return false
 	}


drakkan:
Thanks, but we don’t need assistance for a simple change like this, opening an issue is enough. Has this been tested, or is it AI-generated code that hasn’t been verified?

philipp-holler:
The code is AI-generated and I tested it by running SFTPGo locally and verifying that a copy/move works with a real Azure Blob Storage backend. Also, all good if you just want to do the implementation yourself, feel free to close the PR in that case 👍

drakkan:
Thank you

PR #2263, opened 2026-07-11
add IAM auth option to postgres, closes #2262

### LINKED ISSUE #2262: RDS IAM auth
### Is your feature request related to a problem? Please describe.

We are using RDS IAM auth, but sftpgo only supports passwords. With IAM auth, token/password is generated via [rds build-auth-token](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/rds/auth).
This token is short lived (15 minutes), so it requires to be implemented in `BeforeConnect` in postgres/mysql config.

### Describe the solution you'd like

New option to use IAM auth where I pass AWS region (everything else e.g. host, port etc. can be retrieved from existing flags).

### Describe alternatives you've considered

There is no alternative currently, we can get token on startup, but as soon as sftpgo will acquire new connection after 15 minutes, this connection will be refused.

Medium business

### Additional context

_No response_

pete911:
Initial PR to address #2262

# Checklist for Pull Requests

---

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/aws/aws-sdk-go-v2 v1.42.1
 	github.com/aws/aws-sdk-go-v2/config v1.32.27
 	github.com/aws/aws-sdk-go-v2/credentials v1.19.26
+	github.com/aws/aws-sdk-go-v2/feature/rds/auth v1.6.30
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.104.2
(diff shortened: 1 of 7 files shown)


drakkan:
Thanks for taking the time to open this PR.

I understand the use case, but I don't think this is a good fit for the project as an open-source feature. Supporting RDS IAM auth properly would mean testing and maintaining it against a live AWS account going forward, that's an ongoing cost that doesn't scale for a maintainer-funded project.

PR #2266, opened 2026-07-18
Query OIDC's UserInfo endpoint for authorotative claims

### LINKED ISSUE #2265: [Bug]: Non-compliant OIDC claim reading

### Bug description

I've been chatting with the Authelia maintainer, and so developed the understanding that the SFTPGo's current use of the auth token alone, with no userinfo call, doesn't fully implement the design set out in §5 of the OIDC core spec.

Because this is one of the more common ways that OIDC clients fall short of the spec, many OIDC providers provide settings to work around this (`claims_policies` in Authelia's case). However, it would be much better to just use the OIDC userinfo endpoint as intended.

Also see: https://github.com/drakkan/sftpgo/issues/1493#issuecomment-2832050879

### Steps to reproduce

Use a compliant OIDC provider, try to use info exposed through the userinfo endpoint.

### Expected behavior

SFTPGo calls the userinfo endpoint and reads claims/data from it.

### SFTPGo version

2.7.4

Private user, home usecase (home backup/VPS)

tecosaur:
This is something that's niggled me ever since I set up SFTPGo and discovered it was one of the handful of OIDC clients that hasn't quite faithfully implemented the OIDC spec and needs a workaround.

As per [OIDC Core §5.3](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo), the authoritative way to receive OIDC claims is through the userinfo endpoint. Currently, SFTPGo doesn't query this, and relies on all claims being shoved into the token.

So, I've gone ahead and added userinfo reading to the OIDC client implementation. Specifically, unless `id_token_claims_only` is set (only needed in case of broken OIDC providers), we try to access the userinfo endpoint, and upon a response:
- check that the response is valid, otherwise abort the login (something is wrong)
- merge the userinfo claims with the token's

It goes without saying that auth needs to be handled with care; this isn't the first time I've contributed OIDC improvements. I've upstreamed OIDC implementations/improvements to a handful of projects at this point, and consulted the Authelia maintainer to make sure I had the right understanding of how token/userinfo merging should be handled and what's a decent approach to non-conformant OIDC providers.

I got an LLM to both help with reviewing this work, and also write tests for it.

Closes #2265

## Checklist for Pull Requests

### DIFF
--- a/internal/httpd/httpd_test.go
+++ b/internal/httpd/httpd_test.go
@@ -27681,6 +27681,10 @@ func startOIDCMockServer() {
 			w.WriteHeader(http.StatusNotFound)
 			fmt.Fprintf(w, "Not found\n")
 		})
+		// tests that don't set a UserInfo mock fall back to id-token claims deterministically
+		http.HandleFunc("/auth/realms/sftpgo/protocol/openid-connect/userinfo", func(w http.ResponseWriter, _ *http.Request) {
+			w.WriteHeader(http.StatusUnauthorized)
+		})
 		if err := http.ListenAndServe(oidcMockAddr, nil); err != nil {
 			logger.ErrorToConsole("could not start HTTP notification server: %v", err)
 			os.Exit(1)
--- a/internal/httpd/httpd_test.go
+++ b/internal/httpd/httpd_test.go
@@ -27681,10 +27681,6 @@ func startOIDCMockServer() {
 			w.WriteHeader(http.StatusNotFound)
 			fmt.Fprintf(w, "Not found\n")
 		})
-		// tests that don't set a UserInfo mock fall back to id-token claims deterministically
-		http.HandleFunc("/auth/realms/sftpgo/protocol/openid-connect/userinfo", func(w http.ResponseWriter, _ *http.Request) {
-			w.WriteHeader(http.StatusUnauthorized)
-		})
 		if err := http.ListenAndServe(oidcMockAddr, nil); err != nil {
 			logger.ErrorToConsole("could not start HTTP notification server: %v", err)
 			os.Exit(1)
--- a/internal/httpd/oidc.go
+++ b/internal/httpd/oidc.go
@@ -63,6 +63,12 @@ type OIDCTokenVerifier interface {
 	Verify(ctx context.Context, rawIDToken string) (*oidc.IDToken, error)
 }
 
+// OIDCUserInfoFetcher defines an interface for querying the OpenID UserInfo
+// endpoint, so we can mock it in tests. *oidc.Provider satisfies it.
+type OIDCUserInfoFetcher interface {
+	UserInfo(ctx context.Context, tokenSource oauth2.TokenSource) (*oidc.UserInfo, error)
+}
+
 // OIDC defines the OpenID Connect configuration
 type OIDC struct {
 	// ClientID is the application's ID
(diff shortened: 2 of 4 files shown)


drakkan:
Thanks for the update, the overall structure looks good. I just did a quick review, so I may be wrong on some points, I'm also open to discuss the proposed changes, not strong opinions.

- Please rename the setting to `query_userinfo`. I would also prefer `false` as the default: existing working setups keep working unchanged after upgrading.
- Add env var support for the new setting in `internal/config/config.go`, see for example `SFTPGO_HTTPD__BINDINGS__%v__OIDC__DEBUG`.
- Merge precedence: if the default ends up being `true`, the ID token should win on conflicts. With UserInfo taking precedence and enabled by default, an upgrade may silently change the resolved username_field/role_field values for existing setups: for example a Keycloak deployment with a custom mapper overriding preferred_username with only "Add to ID token" enabled, a possible configuration given that SFTPGo has always read claims from the ID token.
- Also assuming `true` as the default: with custom scopes, the access token may not be valid at the UserInfo endpoint (e.g. Entra ID with scopes for a custom API, rare but possible). Those deployments will get login failures after upgrading until they set `query_userinfo` to false. The endpoint gating doesn't help here, since the endpoint is advertised but rejects the token. At a minimum this needs to be clearly documented in the release notes (If I recall about this :smile:)

tecosaur:
Thanks for giving this a look at your thoughts. I'll be able to act on your feedback in a couple of days.

> Please rename the setting to `query_userinfo`

:+1: 

> Add env var support for the new setting in `internal/config/config.go`, see for example `SFTPGO_HTTPD__BINDINGS__%v__OIDC__DEBUG`.

:+1: thanks for the pointer

> I would also prefer `false` as the default

From a "default to unchanged" behaviour here, I absolutely see why this is what comes to mind as the most conservative approach. It just does very much seem to me like the wrong default, from my understanding of the flow OIDC Core seems to encourage (which may well be incorrect/incomplete).

Considering the providers that I know of/have taken a peek at (as mentioned in the issue)/seen docs on, it looks like reading userinfo by default when advertised:
- Causes more correct behaviour when the provider has meaningful extra claims in the userinfo
- Has no impact when the userinfo simply provides the bare minimum (as Microsoft Entra ID seems to)

I think the only possible scenarios where reading userinfo could cause an issue is when:
- Userinfo is advertised but returns something invalid, or
- Userinfo values conflict with token values

These sound like instances of a provider that would fail OpenID certification and/or are misconfigured, i.e. instances when it's probably actually good to fail.

@james-d-elliott it would be useful to hear if this sounds like the right interpretation/assesment of the situation to you?

drakkan:
@tecosaur For SFTPGo Enterprise, I'll be conservative because I don't want to break existing deployments, and the codebase is already different anyway. 

For the community edition here, I'm fine with enabling the userinfo endpoint by default. These changes will be included in the next major release, so a potential breaking change is acceptable. I just need to remember to mention it in the release notes.

The overall implementation looks fine but I need more time to properly review. Thank you!

tecosaur:
Thanks for finishing the work off @drakkan! I hope the WIP version here was of some help :)

drakkan:
> Thanks for finishing the work off @drakkan! I hope the WIP version here was of some help :)

Sure, it helped, and thanks for reporting this. I made a few changes to the defaults: I inverted the claim precedence and made the feature disabled by default. It should be fully functional with Authelia and the UserInfo endpoint can be used

PR #2295, opened 2026-09-21
Add pbkdf2-sha256 as a FIPS 140 approved password hashing algorithm

### LINKED ISSUE #2294: Feature request: FIPS 140 approved password hashing algorithm (pbkdf2-sha256)
## Is your feature request related to a problem?

SFTPGo currently creates password hashes using only two algorithms
(`password_hashing.algo`): `bcrypt` and `argon2id`. Neither bcrypt nor
argon2id is a NIST/FIPS 140 approved algorithm. This makes it impossible
to build a FIPS-compliant SFTPGo image or run SFTPGo in an environment
that requires FIPS 140-validated cryptographic modules end-to-end,
since there is no way to disable or avoid the non-approved hashing
algorithms.

This is needed to deploy SFTPGo consistently across FedRAMP and
non-FedRAMP clusters using the same image/config, rather than
maintaining a divergent build for FedRAMP environments.

## Describe the solution you'd like

Add a NIST/FIPS-approved option for `password_hashing.algo`, e.g.
`pbkdf2-sha256` (PBKDF2-HMAC-SHA256, per NIST SP 800-132), so
FIPS-conscious deployments can select it instead of bcrypt/argon2id.

Note that SFTPGo already implements PBKDF2 verification for migrating
password hashes from other systems (`comparePbkdf2PasswordAndHash` in
`internal/dataprovider/dataprovider.go`, supporting the
`$pbkdf2-b64salt-sha256$`, `$pbkdf2-sha256$`, `$pbkdf2-sha512$` and
`$pbkdf2-sha1$` hash formats) - it's just never exposed as a hash
*creation* option. Reusing that existing verification code to also
support creating pbkdf2-sha256 hashes should be a small, low-risk
change.

I've opened a PR implementing this: #2295

## Describe alternatives you've considered

- `GOEXPERIMENT=boringcrypto` / a FIPS-enabled Go toolchain: doesn't
  help, since bcrypt and argon2id aren't in the FIPS-approved algorithm
  set and don't route through a FIPS module regardless of build flags.
- Maintaining a private fork with this change: possible under AGPLv3,
  but avoidable if this is accepted upstream, and better for everyone
  who hits the same FIPS requirement.

iAbhishek91:
## Summary

Closes #2294.

`password_hashing.algo` currently only supports `bcrypt` and
`argon2id`, neither of which is a NIST/FIPS 140 approved algorithm.
This prevents building a FIPS-compliant SFTPGo image, since there's
no FIPS-approved algorithm to select even with a FIPS-enabled Go
toolchain.

This PR adds `pbkdf2-sha256` (PBKDF2-HMAC-SHA256, per NIST SP 800-132)
as a third selectable value for `password_hashing.algo`, plus a new
`password_hashing.pbkdf2_options.iterations` setting (default
`600000`, minimum enforced `10000` per NIST SP 800-132).

SFTPGo already implements PBKDF2 verification, used for migrating
password hashes from other systems (`comparePbkdf2PasswordAndHash`,
supporting `$pbkdf2-b64salt-sha256$`, `$pbkdf2-sha256$`,
`$pbkdf2-sha512$`, `$pbkdf2-sha1$`). This change reuses that existing
verification path for the new `pbkdf2-sha256` creation option, instead
of adding a parallel implementation.

## Changes

- `internal/dataprovider/dataprovider.go`: add `HashingAlgoPBKDF2SHA256`
  constant and `Pbkdf2Options` config struct; `hashPlainPassword` now
  creates a `$pbkdf2-b64salt-sha256$<iterations>$<salt>$<hash>` hash
  when selected, using `util.GenerateRandomBytes` for the salt and the
  existing `pbkdf2SHA256B64SaltPrefix` format so it round-trips through
  the existing `comparePbkdf2PasswordAndHash` verifier unchanged;
  `initializeHashingAlgo` validates the configured iteration count.
- `internal/config/config.go`: default `pbkdf2_options.iterations` to
  `600000` and register the corresponding viper default.
- `internal/dataprovider/admin.go`, `apikey.go`, `share.go`: these
  duplicated the bcrypt/argon2id branching for hashing admin passwords,
  API keys and share passwords. Replaced with a call to the shared
  `hashPlainPassword` helper so all four credential types pick up
  `pbkdf2-sha256` consistently, and added the matching
  `comparePbkdf2PasswordAndHash` branch to each type's verification
  path (previously only `dataprovider.go`'s user-password verification
  handled pbkdf2 hashes; admin/API key/share verification assumed
  bcrypt-or-argon2id).

This is intentionally scoped to just the hashing algorithm; it doesn't
touch transport/TLS crypto, which is a separate FIPS concern already
addressed by building with a FIPS-enabled Go toolchain
(`GOEXPERIMENT=boringcrypto`), as noted in #1272.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./internal/dataprovider/... ./internal/config/...`
- [x] `gofmt -l` on all changed files (no output)
- [x] Manually verified hash/verify round-trip for `pbkdf2-sha256`
      (correct password matches, wrong password is rejected, hash has
      the expected `$pbkdf2-b64salt-sha256$` format) with a local test;
      removed before submitting since this package has no existing
      test file to extend
- [ ] Would appreciate maintainer guidance on whether a
      `dataprovider_test.go` should be introduced for this, since none
      currently exists in this package


drakkan:
see https://github.com/drakkan/sftpgo/issues/2294#issuecomment-5781623497. Thanks anyway

